PR #2, opened 2025-12-22
docs: clarify Homebrew vs local build installation steps

nitinalabur:
Update README Install / Build section to explicitly state that Homebrew installation and local build are alternative options and clarify that the run command is for local build only.


steipete:
thanks!

PR #10, opened 2026-01-26
feat(send): add --reply-to support for quoted replies

Dramex:
Adds quoted reply support for text sends.

New flags:
- --reply-to <msgId> (stanza id)
- --reply-chat <jid> (optional; defaults to --to)

Implementation:
- Adds wa.Client.SendTextReply using ContextInfo (StanzaID, Participant, QuotedMessage)
- Extends app.WAClient interface
- Updates send text command to build quoted message from local DB

Notes:
- For group replies, participant is derived from the quoted message sender when available.
- If the reply-to message is not in the local DB, the command suggests running sync/backfill.

Tests: go test ./...

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,11 @@ pnpm wacli history backfill --chat 1234567890@s.whatsapp.net --requests 10 --cou
 # Send a message
 pnpm wacli send text --to 1234567890 --message "hello"
 
+# Reply to a message (quoted reply)
+# Note: for group replies, wacli needs the original sender JID (participant).
+# Make sure the message exists in your local DB (run `wacli sync --follow` or `wacli history backfill`).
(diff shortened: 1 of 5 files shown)


vlassance:
Great to see this\! I submitted a similar implementation in #68 that also covers `send file` (with `ContextInfo` on media messages) and adds a `self` shorthand for `--reply-to-participant` so you don't need to know your own JID. Your approach of looking up the quoted message from the local DB to populate `QuotedMessage` is really nice for the UX — clients that render quoted replies will show the original text. Would be good to combine the best of both.

steipete:
Closing as superseded by #154 / 782c290, which added quoted text replies. Thanks @mmazzarolo.

PR #19, opened 2026-01-30
fix: resolve individual sender JID in LID-addressed groups

### LINKED ISSUE #18: Group messages: SenderJID shows group JID instead of individual participant
When listing or searching messages in a group chat, `SenderJID` is always the group JID for all non-me messages, making it impossible to attribute messages to individual participants.

## Reproduction

```bash
wacli messages list --chat <group-jid> --limit 5 --json
```

Every message returns `"SenderJID":"<group-jid>"` regardless of who sent it. Only `FromMe` distinguishes own messages.

## Expected

`SenderJID` should resolve to the individual participant's JID (or LID), so messages can be attributed to specific people.

## Likely cause

The group uses LID addressing mode (`"AddressingMode":"lid"` in `groups info --jid <jid> --json`). Participant LIDs are available in group info but aren't being mapped to individual messages during storage or retrieval.

## Workaround

Cross-reference `groups info --jid <jid> --json` to get the participant list — but still can't tell who sent which message.

## Environment

- wacli 0.2.0
- Linux x86_64 (Homebrew install)

entropyy0:
Closes #18

In LID-addressed groups, `key.participant` is empty, so `ParseHistoryMessage` falls back to `key.remoteJID` (the group JID). All messages appear to come from the group itself.

The WhatsApp proto has two participant fields:
- `WebMessageInfo.key.participant` (field 4) — empty in LID groups
- `WebMessageInfo.participant` (field 5) — contains the individual sender

Fix: check top-level `hist.GetParticipant()` first, matching the priority used by whatsmeow's own `ParseWebMessage` (client.go:940-943).
Two tests added, full suite passes. AI-assisted (Claude), fully tested.

### DIFF
--- a/internal/wa/messages.go
+++ b/internal/wa/messages.go
@@ -65,7 +65,13 @@ func ParseHistoryMessage(chatJID string, hist *waProto.WebMessageInfo) ParsedMes
 		FromMe:    hist.GetKey().GetFromMe(),
 	}
 
-	sender := strings.TrimSpace(hist.GetKey().GetParticipant())
+	// Prefer top-level participant (works for LID-addressed groups),
+	// then fall back to key.participant, then key.remoteJID.
+	// This matches whatsmeow's own ParseWebMessage priority.
+	sender := strings.TrimSpace(hist.GetParticipant())
+	if sender == "" {
+		sender = strings.TrimSpace(hist.GetKey().GetParticipant())
+	}
 	if sender == "" {
 		sender = strings.TrimSpace(hist.GetKey().GetRemoteJID())
 	}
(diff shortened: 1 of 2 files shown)


vlassance:
This tackles the sender JID side of the LID problem — nice. I submitted #66 which handles the chat JID side: resolving LID-based chat JIDs to phone number JIDs at the storage layer using `Store.LIDs.GetPNForLID()`. Together these would cover both the chat-level and sender-level LID resolution. They're complementary.

steipete:

steipete:
Closing as already landed on main in e73dba0 (#19). Thanks @entropyy0.

PR #20, opened 2026-01-30
feat: Unix socket IPC for send while sync is running

sergi039:
## Problem

`wacli send` cannot work while `sync --follow` is running because both commands need an exclusive lock on the store directory. This makes it impossible to send messages programmatically while maintaining a sync session.

## Solution

`sync --follow` now starts a Unix socket server that accepts send commands from other wacli processes.

### How it works:
1. `sync --follow` starts an IPC server at `~/.wacli/wacli.sock`
2. `send` checks if the socket exists and tries IPC first
3. If IPC succeeds, message is sent through the running sync daemon
4. If IPC fails or socket doesn't exist, falls back to direct mode

## Changes

- `internal/ipc/ipc.go`: New IPC server/client using Unix sockets
- `cmd/wacli/sync.go`: Start IPC server in `--follow` mode (`--enable-ipc=true` by default)
- `cmd/wacli/send.go`: Try IPC first, fallback to direct if unavailable

## Usage

```bash
# Terminal 1: Start sync daemon
wacli sync --follow

# Terminal 2: Send messages (routes through daemon automatically)
wacli send text --to 1234567890 --message 'Hello!'
# Output: Sent to 1234567890@s.whatsapp.net (id XXX) via daemon
```

## New Flags

- `sync --enable-ipc=false` - disable IPC server
- `send --no-ipc` - skip IPC, use direct mode only

## Testing

Successfully tested sending messages via IPC while sync was running.

### DIFF
--- a/cmd/wacli/delete.go
+++ b/cmd/wacli/delete.go
@@ -1,0 +1,120 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+	"github.com/steipete/wacli/internal/config"
+	"github.com/steipete/wacli/internal/ipc"
+	"github.com/steipete/wacli/internal/out"
+	"github.com/steipete/wacli/internal/wa"
+)
+
+func newDeleteCmd(flags *rootFlags) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delete",
+		Short: "Delete messages",
+	}
+	cmd.AddCommand(newDeleteMessageCmd(flags))
+	return cmd
+}
+
+func newDeleteMessageCmd(flags *rootFlags) *cobra.Command {
+	var chat string
+	var msgID string
+	var forEveryone bool
+	var noIPC bool
+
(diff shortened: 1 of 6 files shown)


vlassance:
Nice work on this! I took a similar approach in #65 — a Unix socket server started during `sync --follow`. A few differences in my implementation: I kept the server in `internal/app/sendserver.go` instead of a separate `ipc` package, and I added file send support (image/video/audio/document with MIME detection) alongside text. Your approach of including a `delete` command and the `--enable-ipc` / `--no-ipc` flags is a nice touch. Either way, glad to see this problem getting attention.

dinakars777 (commented):
Thanks for exploring the send-while-sync IPC path. I’m going to close this PR because this area now has multiple overlapping implementations, and we need to consolidate around one current design before merging anything.

This PR is an older version of the Unix-socket delegation approach. Keeping several competing IPC PRs open makes review harder and increases the chance that we merge inconsistent behavior.

I’m closing this as superseded by the newer IPC/send-while-sync work rather than merging this specific implementation.

PR #22, opened 2026-02-01
Add --qr-format flag for raw QR payload output

teren-papercutlabs:
## Summary

Adds a `--qr-format` flag to `wacli auth` that controls QR code output format:
- `terminal` (default): Current behavior - ASCII art via qrterminal
- `text`: Raw QR payload string to stdout

## Use Case

When building web-based WhatsApp onboarding flows, the raw QR payload can be rendered with a proper QR library (like `qrcode` in Node.js) to guarantee a square, scannable QR code. The ASCII art approach requires parsing half-height block characters (`▀`/`▄`) which is fragile and can produce rectangular output.

## Changes

- Added `--qr-format` string flag to `auth` command (default: "terminal")  
- When `--qr-format=text`, outputs raw QR string to stdout instead of rendering ASCII art

## Testing

```bash
wacli auth --qr-format=text --store /tmp/test
# Outputs: 2@abc123...  (raw QR payload)
```

### DIFF
--- a/cmd/wacli/auth.go
+++ b/cmd/wacli/auth.go
@@ -18,6 +18,7 @@ func newAuthCmd(flags *rootFlags) *cobra.Command {
 	var follow bool
 	var idleExit time.Duration
 	var downloadMedia bool
+	var qrFormat string
 
 	cmd := &cobra.Command{
 		Use:   "auth",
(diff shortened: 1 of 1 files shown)


r4jiv007:
this will solve the same problem i am facing.

@teren-papercutlabs  also if you can read the json flag and print in json format, that would be good.

something like 
`{ "code" :" ......" }`

steipete:
Closing as superseded by a focused main patch: `eabf8d6` (`feat: print raw auth QR payload`).

What landed:
- `wacli auth --qr-format text` prints the raw WhatsApp QR payload to stdout.
- Default remains `--qr-format terminal`, preserving the existing terminal QR behavior.
- Unsupported formats fail clearly.
- `--qr-format=text` is rejected with `--json` so stdout cannot contain mixed raw QR text and final JSON.

Dependency proof: current whatsmeow `QRChannelItem.Code` is documented in source as the raw QR data emitted for `QRChannelEventCode`, so this is just exposing the dependency-provided payload without parsing terminal QR art.

Verified:
- `go test ./cmd/wacli -run 'TestNormalizeAuthQRFormat|TestAuthQRWriterText|TestAuthCommandExposesQRFormat'`

Thanks for the PR; I skipped the committed binary from the branch and landed the source-only version.

PR #45, opened 2026-02-17
fix: resolve LID senders to contact names in message output

### LINKED ISSUE #18: Group messages: SenderJID shows group JID instead of individual participant

decodiver22:
## Summary

Fixes #18 — WhatsApp groups using LID addressing showed anonymized LIDs (e.g. `248317879549994@lid`) instead of real sender names in `messages list/search/show/context`.

**Root cause:** Two issues:
1. **Sync-time:** `storeParsedMessage` called `GetContact(jid.ToNonAD())` on LID senders, but `ToNonAD()` on a LID still yields a LID — contacts are stored under phone numbers, so the lookup always failed.
2. **Display-time:** The `store.Message` struct had no `SenderName` field, so even when `sender_name` was written to the DB, it was never read back.

**Fix:**
- Read `sender_name` from DB in all message queries (`ListMessages`, `GetMessage`, `MessageContext`, `SearchMessages`)
- Resolve LID → phone number via `store.LIDs.GetPNForLID` at sync time before contact lookup
- Add `ResolveSenderNames` helper for display-time resolution of remaining unresolved LIDs
- All 4 message subcommands call resolution and prefer `SenderName` over raw `SenderJID` in output

## Changes

- `internal/store/types.go` — Add `SenderName` to `Message` struct
- `internal/store/messages.go` — Read `sender_name` in all SELECT queries + scan
- `internal/store/search.go` — Same for search queries
- `internal/wa/client.go` — New `GetPNForLID` method
- `internal/app/app.go` — Extend `WAClient` interface
- `internal/app/resolve.go` — **New**: `ResolveSenderNames` display-time helper
- `internal/app/sync.go` — LID→PN resolution before contact lookup at sync
- `cmd/wacli/messages.go` — Call resolution, show names in output
- `internal/app/resolve_test.go` — **New**: 7 test cases for resolution logic
- `internal/app/fake_wa_test.go` — Add `GetPNForLID` to mock

## Test plan

- [x] `go vet ./...` — clean
- [x] `go test ./...` — all pass
- [x] `go test -tags sqlite_fts5 ./...` — all pass
- [x] Smoke-tested `messages list` on LID groups — senders show real names
- [x] Smoke-tested `messages search/show/context` — names resolved
- [x] DM regression — no change in behavior
- [x] JSON output — `SenderName` field present, `SenderJID` preserved (backwards-compatible)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.5.0 - Unreleased
 
+### Fixed
+
+- Messages: resolve LID senders in group chats to real contact names or phone numbers (#18).
+
 ### Changed
 
 - Internal architecture: split store and groups command logic into focused modules for cleaner maintenance and safer follow-up changes.
--- a/cmd/wacli/messages.go
+++ b/cmd/wacli/messages.go
@@ -71,6 +71,8 @@ func newMessagesListCmd(flags *rootFlags) *cobra.Command {
 				return err
 			}
 
+			a.ResolveSenderNames(ctx, msgs)
+
 			if flags.asJSON {
 				return out.WriteJSON(os.Stdout, map[string]any{
 					"messages": msgs,
@@ -82,6 +84,9 @@ func newMessagesListCmd(flags *rootFlags) *cobra.Command {
 			fmt.Fprintln(w, "TIME\tCHAT\tFROM\tID\tTEXT")
 			for _, m := range msgs {
 				from := m.SenderJID
+				if m.SenderName != "" {
+					from = m.SenderName
+				}
 				if m.FromMe {
 					from = "me"
 				}
@@ -168,6 +173,8 @@ func newMessagesSearchCmd(flags *rootFlags) *cobra.Command {
 				return err
 			}
 
+			a.ResolveSenderNames(ctx, msgs)
+
 			if flags.asJSON {
 				return out.WriteJSON(os.Stdout, map[string]any{
 					"messages": msgs,
@@ -179,6 +186,9 @@ func newMessagesSearchCmd(flags *rootFlags) *cobra.Command {
 			fmt.Fprintf(w, "TIME\tCHAT\tFROM\tID\tMATCH\n")
 			for _, m := range msgs {
 				fromLabel := m.SenderJID
+				if m.SenderName != "" {
+					fromLabel = m.SenderName
+				}
 				if m.FromMe {
 					fromLabel = "me"
 				}
@@ -244,6 +254,10 @@ func newMessagesShowCmd(flags *rootFlags) *cobra.Command {
 				return err
 			}
 
+			msgs := []store.Message{m}
+			a.ResolveSenderNames(ctx, msgs)
+			m = msgs[0]
+
 			if flags.asJSON {
 				return out.WriteJSON(os.Stdout, m)
 			}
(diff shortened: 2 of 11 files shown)


decodiver22:
Note: PR #19 by @entropyy0 also addresses issue #18 with a different approach — patching the proto parsing layer to prefer `WebMessageInfo.participant` (field 5) over `key.participant` (field 4) for LID groups.

This PR takes a complementary approach: instead of fixing the proto parsing, it resolves LIDs to phone numbers using whatsmeow's built-in `store.LIDs.GetPNForLID`, then looks up the contact name. This works for both history sync and live messages, and also handles display-time resolution for messages already stored without a sender name.

Both approaches could potentially be combined — #19 would fix the SenderJID attribution at the proto level, while this PR resolves LIDs to human-readable names.

vlassance:
This solves the display/output side nicely. I submitted #66 which resolves LID JIDs at the storage layer — so messages are stored under the PN-based chat JID rather than the LID one. These are complementary: yours fixes how LID senders are shown, mine fixes where LID messages are stored.

steipete:
Partial maintainer-owned pieces from this landed on main:

- `fca5b96` resolves live LID chat/sender JIDs to PN JIDs before storing new sync messages.
- `4481fc8` expands `messages list/search --chat <phone>` through the whatsmeow PN->LID map so old split `@lid` chat rows are visible.
- `02f98c3` reads the already-stored `sender_name` column back from SQLite and prefers it in message output.
- `87095bd` applies the same mapped chat lookup to `messages show/context`.

I verified the dependency contract against whatsmeow `Store.LIDs.GetPNForLID` / `GetLIDForPN` and ran the full local gate after the patches:
`pnpm format:check && pnpm lint && pnpm test && pnpm build && git diff --check`

Not closing this PR yet because one part of the PR is still a distinct behavior: display-time resolution for historical rows that have `sender_jid=@lid` and an empty `sender_name`. The current main fixes new writes and stored sender-name reads, but it does not yet synthesize a sender name for every old blank row.

steipete:
Closing as superseded by maintainer-owned fixes now on main.

The PR identified the right bug class. I split and landed the current-main version in focused commits:
- `fca5b96` resolves live LID chat/sender JIDs before storage.
- `4481fc8` expands `messages list/search --chat <phone>` through whatsmeow LID mappings.
- `02f98c3` reads stored `sender_name` back from SQLite and prefers it in output.
- `87095bd` applies mapped chat lookup to `messages show/context`.
- `787cfd5` resolves historical blank `@lid` sender names at display time using whatsmeow `GetPNForLID` plus local contacts.

Verified with the full local gate after the landed patches:
`pnpm format:check && pnpm lint && pnpm test && pnpm build && git diff --check`

Thanks for the detailed root-cause writeup here; it made the final split much easier to land safely.

PR #64, opened 2026-02-19
feat: track message edits, revokes, and delete-for-me events

vlassance:
## Summary

- Add schema migration v4 with `edit_type`, `edit_target_id`, `updated_at`, `reaction_to_id`, and `reaction_emoji` columns on the messages table
- Detect message edits via `ProtocolMessage_MESSAGE_EDIT` and update the original message in-place
- Extract target message ID from `ProtocolMessage_REVOKE` for sender-deleted messages
- Handle `events.DeleteForMe` by inserting synthetic revoke rows (prefixed `dfm-`)
- Store structured reaction metadata (`reaction_to_id`, `reaction_emoji`) alongside the existing `display_text`
- Auto-update `updated_at` via trigger on message modifications

## Problem

wacli currently has no way to track when messages are edited, revoked (deleted for everyone), or deleted for the current user. This makes it impossible to maintain an accurate message history or detect message modifications.

## Solution

Add new columns to the messages table via the existing versioned migration system (v4). Message edits update the original row in-place while also storing the edit event. Revoke and delete-for-me events are stored as synthetic rows linked to the target message via `edit_target_id`.

## Testing

### DIFF
--- a/internal/app/sync.go
+++ b/internal/app/sync.go
@@ -95,14 +95,27 @@ func (a *App) Sync(ctx context.Context, opts SyncOptions) (SyncResult, error) {
 					}
 				}
 			}
-			if err := a.storeParsedMessage(ctx, pm); err == nil {
-				messagesStored.Add(1)
+			// Handle message edits: update the original message in-place.
+			if pm.EditTargetID != "" && pm.EditType == "1" {
+				displayText := a.buildDisplayText(ctx, pm)
+				_ = a.db.UpdateMessageEdit(pm.Chat.String(), pm.EditTargetID, pm.Text, displayText)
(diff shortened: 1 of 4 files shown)


steipete:
Thanks @vlassance. I rebuilt and landed the delete-for-me part on main in 31504a8, with the schema/tombstone path, `regular_high` app-state catch-up, docs/changelog, tests, and a live WhatsApp proof.

I did not merge this PR as-is because the synthetic dfm row/edit-target schema overlapped with newer main behavior and would have introduced a second representation for deleted messages. The shipped implementation keeps delete-for-me as an in-place tombstone on the original message row instead. Closing this as superseded by main.

PR #65, opened 2026-02-19
feat: Unix socket IPC for send while sync is running

### LINKED ISSUE #6: How about allowing sending while sync is running?
Ideally having `wacli sync` running wouldn't block `wacli send`, so you could keep a long-running process up-to-date... sure you could script up something to stop the sync, send a message, and resume syncing, but it seems like it'd be a common enough use case to build in directly, no?

vlassance:
## Summary

- Start a Unix socket server at `<store>/send.sock` when `sync --follow` is running
- `send text` and `send file` try the socket first, falling back to direct connection
- Supports text messages and file uploads (image/video/audio/document) with MIME detection
- Socket uses JSON-over-newline protocol with 30s timeout (2min for file uploads)
- Socket file is created with 0600 permissions and cleaned up on shutdown

## Problem

`wacli send` cannot work while `sync --follow` is running because both commands need the store. This makes it impossible to send messages programmatically while maintaining a sync session (see #6).

## Solution

`sync --follow` now starts a Unix socket server that accepts send requests from other wacli processes. The `send` commands check for the socket first and route through the running sync daemon, falling back to a direct connection if sync is not running.

## How it works

1. `sync --follow` starts a socket server at `~/.wacli/send.sock`
2. `send text` / `send file` check if the socket exists and try IPC first
3. If IPC succeeds, the message is sent through the running sync daemon's connection
4. If IPC fails or socket doesn't exist, falls back to direct mode

## Protocol

Each request/response is a single JSON line terminated by `\n`:

```json
{"to": "+1234567890", "message": "hello"}
{"to": "+1234567890", "type": "file", "file_path": "/tmp/photo.jpg", "caption": "check this out"}
```

## Testing

Closes #6

### DIFF
--- a/cmd/wacli/send.go
+++ b/cmd/wacli/send.go
@@ -1,12 +1,18 @@
 package main
 
 import (
+	"bufio"
 	"context"
+	"encoding/json"
 	"fmt"
+	"net"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/spf13/cobra"
+	"github.com/steipete/wacli/internal/app"
+	"github.com/steipete/wacli/internal/config"
 	"github.com/steipete/wacli/internal/out"
 	"github.com/steipete/wacli/internal/store"
 	"github.com/steipete/wacli/internal/wa"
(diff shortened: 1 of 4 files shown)


dinakars777 (commented):
Thanks for the focused IPC work here. I’m going to close this PR because it overlaps heavily with the broader send-while-sync IPC direction, and we need to avoid carrying multiple competing socket protocols and file locations.

The core problem is valid, but we should land one current, reviewed implementation rather than merge parallel versions with slightly different behavior.

Closing this as superseded/consolidated into the newer IPC direction.

Issue #28, opened 2026-02-03
Phone numbers with + prefix cause 'usync query timed out' error


miguelalcalde:
## Summary

When using `wacli send` with a phone number that includes the `+` prefix (e.g., `+49123456789`), the command fails with "info query timed out". Removing the `+` prefix makes it work instantly.

## Steps to Reproduce

```bash
# This times out after ~80 seconds:
wacli send text --to "+49123456789" --message "test"

# This works instantly:
wacli send text --to "49123456789" --message "test"
```

## Expected Behavior

Both formats should work. The `+` prefix should be normalized/stripped before building the JID for the usync query.

## Actual Behavior

```
$ wacli send text --to "+49123456789" --message "test"
failed to get user info for +49123456789@s.whatsapp.net to fill LID cache: failed to send usync query: info query timed out
```

Note the malformed JID: `+49123456789@s.whatsapp.net` — the `+` is included when it should be stripped.

## Environment

- wacli version: 0.2.0
- OS: macOS 12.x (Darwin 21.6.0, x86_64)

## Workaround

Strip the `+` prefix before passing to wacli, or use full JID format: `49123456789@s.whatsapp.net`

## Notes

- `wacli chats list` and other read operations work fine
- Sending to known JIDs directly works fine
- Only the phone number → JID lookup path is affected

vlassance:
This may be related to WhatsApp's LID migration — some JID lookups fail when the phone-to-LID mapping isn't resolved. #66 adds LID-to-PN resolution at the storage layer which helps with the broader JID consistency problem. It won't fix the `+` prefix parsing directly, but it does address cases where LID-based JIDs cause unexpected behavior.

PR #66, opened 2026-02-19
fix: resolve LID JIDs to phone number JIDs

vlassance:
## Summary

- Add `ResolveLIDToPN()` to `WAClient` interface and `wa.Client`
- Resolve Linked Identity (LID) JIDs to Phone Number JIDs when storing messages
- Prevents the "split chat" issue caused by WhatsApp's LID migration

## Problem

WhatsApp is migrating to Linked Identity (LID) JIDs (with server `lid`), which causes messages to be stored under separate chat entries instead of the canonical phone-number-based chat. This results in "split chats" where the same conversation appears as two different chats in the local database (see #18, #28).

## Solution

Before storing a message, resolve any LID-based chat JID to its corresponding Phone Number JID using the whatsmeow store's `LIDs.GetPNForLID()`. If resolution fails (e.g., the LID is unknown), the original JID is kept unchanged.

## Testing

Relates to #18, #28

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -26,6 +26,7 @@ type WAClient interface {
 	ReconnectWithBackoff(ctx context.Context, minDelay, maxDelay time.Duration) error
 
 	ResolveChatName(ctx context.Context, chat types.JID, pushName string) string
+	ResolveLIDToPN(ctx context.Context, jid types.JID) types.JID
 	GetContact(ctx context.Context, jid types.JID) (types.ContactInfo, error)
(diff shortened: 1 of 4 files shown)


steipete:
Thanks @vlassance. This is now superseded by fca5b96 on main.

That maintainer patch resolves live @lid chat/sender JIDs through whatsmeow own LID store before storage, adds a regression test, and passed the full local gate. I am closing this older branch since it is dirty and covered by the landed fix.

PR #67, opened 2026-02-19
feat: store structured reaction data (reaction_to_id, reaction_emoji)

vlassance:
## Summary

- Add `reaction_to_id` and `reaction_emoji` columns to the messages table via migration v4
- Pass structured reaction data through `UpsertMessageParams` and into the database
- Complements the `display_text` approach from PR #5 with machine-readable reaction data

## Problem

PR #5 added `display_text` for reactions (e.g., "Reacted 👍 to hello"), which is great for human-readable output. However, there is no way to programmatically query which message a reaction targets or what emoji was used. This makes it difficult to build downstream tools that need to process reactions (e.g., counting reactions, filtering by emoji).

## Solution

Store the raw `reaction_to_id` (stanza ID of the message being reacted to) and `reaction_emoji` (the emoji text) as separate columns. These are populated from the existing `ParsedMessage.ReactionToID` and `ParsedMessage.ReactionEmoji` fields that are already parsed by the `wa` package.

## Testing

### DIFF
--- a/internal/app/sync.go
+++ b/internal/app/sync.go
@@ -326,6 +326,8 @@ func (a *App) storeParsedMessage(ctx context.Context, pm wa.ParsedMessage) error
 		FileSHA256:    fileSha,
 		FileEncSHA256: fileEncSha,
 		FileLength:    fileLen,
+		ReactionToID:  pm.ReactionToID,
+		ReactionEmoji: pm.ReactionEmoji,
 	})
 }
 
--- a/internal/store/messages.go
+++ b/internal/store/messages.go
@@ -25,6 +25,8 @@ type UpsertMessageParams struct {
 	FileSHA256    []byte
 	FileEncSHA256 []byte
 	FileLength    uint64
+	ReactionToID  string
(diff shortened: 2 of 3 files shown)


steipete:
Closing as superseded by focused main patch: `70e5a20` (`feat: store structured reaction metadata`).

What landed:
- Adds `reaction_to_id` and `reaction_emoji` to the core schema plus migration v6.
- Carries `wa.ParsedMessage.ReactionToID` / `ReactionEmoji` through `storeParsedMessage` into SQLite.
- Reads those fields back through list/search/show/context query scans and JSON output.

Proof:
- Current parser already extracts reaction target/emoji from `ReactionMessage` / `EncReactionMessage` paths; this patch fixes the persistence drop.
- Added schema, store round-trip, and sync regression coverage.
- Verified normal and FTS paths plus full gate:
`pnpm format:check && pnpm lint && pnpm test && pnpm build && git diff --check`

Thanks for the clean PR; main had moved through later schema migrations, so I landed the current migration/scan version directly.

PR #68, opened 2026-02-19
feat: add --reply-to support for send text and send file

### LINKED ISSUE #16: Feature request: Support for mentions (@user) and reply-to in send text
### Current Behavior
`wacli send text` only supports `--to` and `--message` flags.

### Desired Behavior
Add support for:

1. **Mentions** - Ability to mention users in group messages
```bash
wacli send text --to "GROUP_JID" --message "Hey check this" --mention "USER_JID"
```

2. **Reply/Quote** - Ability to reply to a specific message
```bash
wacli send text --to "GROUP_JID" --message "I agree!" --reply-to "MESSAGE_ID"
```

### Use Case
Building an AI assistant that can participate naturally in WhatsApp group conversations - needs to be able to reply to specific messages and mention users to keep context clear.

### Technical Notes
whatsmeow supports both features:
- `SendMessage` with `ContextInfo.QuotedMessage` for replies
- `ContextInfo.MentionedJid` for mentions

Would be happy to help with a PR if you can point me to the right direction!

Thanks for this great tool 🙏

vlassance:
## Summary

- Add `--reply-to` and `--reply-to-participant` flags to `send text` and `send file`
- Use `ContextInfo` with `StanzaID` for quoted replies on both text and media messages
- Support `self` as a special value for `--reply-to-participant` that resolves to own JID
- Add `GetOwnJID()` to `WAClient` interface for JID resolution

## Problem

There is no way to send a quoted reply to an existing message via `wacli send` (see #16). This limits wacli's usefulness for conversational workflows and bot integrations that need to reply in context.

## Solution

Add reply-to flags that construct a `ContextInfo` with the quoted message's stanza ID. For text messages, this wraps the content in an `ExtendedTextMessage`. For media messages, the `ContextInfo` is attached directly to the media proto message (ImageMessage, VideoMessage, AudioMessage, DocumentMessage).

## Usage

```bash
# Reply to a text message
wacli send text --to +1234567890 --message "I agree!" \
  --reply-to ABC123 --reply-to-participant 9876543210@s.whatsapp.net

# Reply with a file
wacli send file --to +1234567890 --file photo.jpg --caption "Here it is" \
  --reply-to ABC123 --reply-to-participant self
```

## Testing

Closes #16

### DIFF
--- a/cmd/wacli/send.go
+++ b/cmd/wacli/send.go
@@ -4,12 +4,15 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/steipete/wacli/internal/out"
 	"github.com/steipete/wacli/internal/store"
 	"github.com/steipete/wacli/internal/wa"
+	waProto "go.mau.fi/whatsmeow/binary/proto"
+	"google.golang.org/protobuf/proto"
 )
 
 func newSendCmd(flags *rootFlags) *cobra.Command {
@@ -25,6 +28,8 @@ func newSendCmd(flags *rootFlags) *cobra.Command {
 func newSendTextCmd(flags *rootFlags) *cobra.Command {
 	var to string
 	var message string
+	var replyTo string
+	var replyToParticipant string
 
 	cmd := &cobra.Command{
 		Use:   "text",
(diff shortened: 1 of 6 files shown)


steipete:
Current main has send text --reply-to with --reply-to-sender inference for groups, so the text half is covered. This PR still adds send file reply support and uses a different participant flag shape, so I am leaving it open rather than calling it superseded.

steipete:
Landed the remaining file-reply part on main in `3ce15af`.

What changed:
- `send file` now accepts `--reply-to` and `--reply-to-sender`, matching the existing `send text` quoted-reply semantics.
- The quoted context is attached to image, video, audio, and document protobuf messages via their upstream `ContextInfo` fields.
- README, docs/spec, and CHANGELOG are updated.

Proof:
- Read the current `whatsmeow` generated protobuf source and confirmed `ImageMessage`, `VideoMessage`, `AudioMessage`, and `DocumentMessage` all expose `ContextInfo`.
- Added regression coverage for reply context construction, `send file` flag exposure, and media-message context attachment.
- Full local gate passed: `pnpm format:check && pnpm lint && pnpm test && pnpm build && git diff --check`.

Thanks for the PR, Victor. Closing this because the useful remaining behavior is now on `main`; I kept the public flag shape aligned with the already-shipped `send text --reply-to-sender` path rather than adding a second participant flag name.

PR #74, opened 2026-02-22
Strip + prefix from phone numbers in ParseUserOrJID

### LINKED ISSUE #28: Phone numbers with + prefix cause 'usync query timed out' error

FrederickStempfle:
Phone numbers with a `+` prefix (like `+49123456789`) get passed straight
into the JID as `+49123456789@s.whatsapp.net`, which is invalid and causes
usync to time out. Just strips the `+` before building the JID.

Added a test case for it too.

Fixes #28

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Media: recover panics per download job so one bad payload no longer drains the worker pool. (#179 — thanks @shaun0927)
 - Messages: show display text for replies, reactions, and media in `messages context`. (#183 — thanks @fuleinist)
+- Send: strip a leading `+` from phone-number recipients before building WhatsApp JIDs. (#74 — thanks @FrederickStempfle)
 - Search: keep FTS5 enabled after reopening existing databases with already-applied migrations. (#185 — thanks @iamhitarth)
 - Send: persist retry-message plaintext so linked devices can decrypt retried messages. (#186 — thanks @SimDamDev)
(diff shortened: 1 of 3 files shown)


steipete:
Landed via rebase in 3e17ff0ae2b728aefdba9a2d51eaf1831c7f0ac3. Expanded the test matrix to cover spaces, explicit JIDs, and `+` by itself; full gate passed locally before push. Thanks @FrederickStempfle.

PR #75, opened 2026-02-22
Filter left groups from groups list

### LINKED ISSUE #23: groups list: filter out groups user has left
## Problem

`wacli groups list` returns all groups ever known to the local DB, including groups the user has already left via `wacli groups leave`. After leaving 17 groups, the list actually grew (from 28 to 48) because `groups refresh` pulled in even more historical groups.

This makes `groups list` unreliable for answering "what groups am I currently in?"

## Suggestion

Either:
- **Default** to showing only currently-joined groups (breaking change, but arguably the correct behavior)
- Add a `--joined` / `--active` flag to filter to current memberships only
- Add a `--left` flag to show only left groups (for cleanup purposes)

`groups refresh` already fetches live membership, so the joined/left state should be available in the local DB.

## Steps to reproduce

```
wacli groups leave --jid <some-group>@g.us   # succeeds
wacli groups list                              # still shows the left group
wacli groups refresh                           # pulls in even more historical groups
wacli groups list                              # list is now larger than before
```

## Environment

- wacli 0.2.0 (Homebrew, macOS)

FrederickStempfle:
`groups list` was returning every group the DB had ever seen, including
ones you already left. After leaving a bunch of groups and running
`groups refresh`, the list actually grew instead of shrinking.

Added an `is_member` column to the groups table (migration 4) so we
can track membership state. `groups leave` now marks the group as left
in the DB, and `UpsertGroup` (called by refresh/sync) resets it back
to joined. `groups list` defaults to showing only joined groups —
pass `--all` to see everything.

Fixes #23

### DIFF
--- a/cmd/wacli/groups_info_rename.go
+++ b/cmd/wacli/groups_info_rename.go
@@ -146,6 +146,7 @@ func newGroupsLeaveCmd(flags *rootFlags) *cobra.Command {
 			if err := a.WA().LeaveGroup(ctx, gjid); err != nil {
 				return err
 			}
+			_ = a.DB().MarkGroupLeft(gjid.String())
 			if flags.asJSON {
 				return out.WriteJSON(os.Stdout, map[string]any{"jid": gjid.String(), "left": true})
 			}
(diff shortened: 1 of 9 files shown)


steipete:
Thanks @FrederickStempfle. Left-group filtering is now on main in 21063d8 and 95f2136, including refresh reconciliation. Closing as superseded.

PR #78, opened 2026-02-23
feat: Refactor CLI commands for testability and better config

nikolasdehor:
This PR refactors the CLI commands to use cmd.OutOrStdout() and cmd.ErrOrStderr() instead of hardcoded os.Stdout/os.Stderr. This enables better testing by allowing output capture in buffers.

Additionally:
- Moves device configuration logic from main.go to internal/config/device.go.
- Adds unit tests for device configuration logic (fixing missing coverage).
- Adds a basic integration test for the version command in cmd/wacli/root_test.go.

This change paves the way for adding more comprehensive CLI tests in the future.

### DIFF
--- a/cmd/wacli/auth.go
+++ b/cmd/wacli/auth.go
@@ -37,7 +37,7 @@ func newAuthCmd(flags *rootFlags) *cobra.Command {
 				mode = appPkg.SyncModeFollow
 			}
 
-			fmt.Fprintln(os.Stderr, "Starting authentication…")
+			fmt.Fprintln(cmd.ErrOrStderr(), "Starting authentication…")
 			res, err := a.Sync(ctx, appPkg.SyncOptions{
 				Mode:            mode,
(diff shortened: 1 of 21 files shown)


steipete:
Thanks for this. I am closing it as superseded rather than trying to rebase the broad refactor.

The concrete useful slice from this PR has already landed on current `main`: `c912668` routes `wacli version` through Cobra's configured output stream and adds the focused regression test. The device-label logic is also covered on main by `cmd/wacli/main_test.go`.

The remaining diff is now a stale repo-wide rewrite of command stdout/stderr plumbing across many commands that have changed heavily since February. Landing that wholesale would create a lot of churn without a current bug to prove, and it would likely conflict with newer command surfaces (`channels`, `profile`, `store`, send delegation, lifecycle events, etc.).

If there is a specific command whose output still cannot be tested or redirected correctly, we should fix that narrowly with a failing test against current `main`.

PR #92, opened 2026-03-11
feat: send messages while sync --follow is running (IPC delegation)

### LINKED ISSUE #70: Store lock prevents sending while sync is running
## Problem

When `wacli sync --follow` is running (continuous listener), attempting to `wacli send text` fails with:

```
store is locked (another wacli is running?): resource temporarily unavailable
```

The only workaround is to stop the sync process, send the message, then restart sync:

```bash
pkill -f "wacli sync" && sleep 2
wacli send text --to "GROUP_JID" --message "hello"
wacli sync --follow &
```

This works but it's clunky, risks missing messages during the gap, and adds complexity for any automation that needs to both listen and send.

## Expected behaviour

Sending a message should work while the sync listener is running. The store lock should either allow concurrent read/write from separate wacli processes, or the sync process could handle sends internally (e.g. via a socket/pipe).

## Environment

- wacli installed via `brew install steipete/tap/wacli`
- macOS 15.3 (Darwin 24.3.0)
- Running `wacli sync --follow` as a background process

najibninaba:
## Summary

Fixes #70.

`wacli send text` and `wacli send file` now work while `sync --follow` is running. When the store lock is held, the send command automatically delegates to the running sync process via a Unix-domain socket (`<store>/.send.sock`). This is fully transparent — output is identical to a direct send.

```bash
# Terminal 1
wacli sync --follow

# Terminal 2 — just works now
wacli send text --to 6591234567 --message "hello"
wacli send file --to 6591234567 --file ./report.pdf --caption "Q3 report"
```

## Design

The fallback chain in each send command:

1. Try to acquire the store lock and send directly (unchanged behavior)
2. If `ErrLocked` → attempt IPC delegation via `.send.sock`
3. If socket unavailable (e.g. `sync --once`) → return the original lock error
4. If the sync process returns an app error (e.g. not authenticated) → surface it
5. If delegation succeeds → format output identically to a direct send

Only `sync --follow` hosts the IPC server. `sync --once` and `auth` do not.

### Architecture

- **Lock sentinel** (`internal/lock`): `ErrLocked` with `errors.Is` support for clean branching
- **Send refactor** (`internal/app/send.go`): `App.SendText`/`SendFile` extracted from CLI layer
- **IPC transport** (`internal/app/send_ipc.go`): JSON-over-UDS protocol (version 1), client + server
- **Sync hosting** (`internal/app/sync.go`): follow-mode starts/stops the IPC server
- **CLI fallback** (`cmd/wacli/send.go`, `send_file_cmd.go`): lock → IPC → original error
- **Concurrency safety**: `connectMu` serializes `Connect`; IPC server serializes delegated sends; file paths normalized to absolute before delegation

### Robustness

- Server-side 1s read deadline prevents stalled clients from blocking future sends
- Active connection tracking + force-close in `Close()` ensures prompt shutdown
- Stale socket cleanup on next startup (crash recovery)
- Socket permissions hardened to `0600`
- File send responses validated for `File != nil` (prevents nil-pointer on malformed peers)

## Commits

| Commit | Description |
|--------|-------------|
| `2d8e812` | `ErrLocked` sentinel + shared `resolveStoreDir` |
| `48e0dc3` | Extract `App.SendText`/`SendFile` into `internal/app/send.go` |
| `056a78c` | IPC transport layer (client + server) |
| `f6e7a57` | Follow-mode sync hosts IPC server + `connectMu` serialization |
| `7ad832a` | CLI send commands: lock → IPC fallback |
| `1958319` | README + spec documentation |
| `cfa8c7c` | Server read deadline + active-conn tracking for robust shutdown |
| `8c3f535` | File response validation + socket permissions `0600` |
| `7a559f3` | Concurrent delegated sends test |

## Test coverage

- Lock contention detection (`lock_test.go`)
- Send behavior: text/file classification, persistence, best-effort DB failures (`send_test.go`)
- IPC round-trip, missing socket, version mismatch, timeout encoding, stale socket cleanup (`send_ipc_test.go`)
- Stalled client does not block shutdown (`send_ipc_test.go`)
- Malformed file response rejected with protocol error (`send_ipc_test.go`)
- Socket permissions = `0600` (`send_ipc_test.go`)
- Follow-mode hosts socket, delegated send succeeds, socket removed on exit (`sync_test.go`)
- Once-mode does not host socket (`sync_test.go`)
- 3 concurrent delegated sends all succeed (`sync_test.go`)
- AST regression: `connectMu` lock discipline in `Connect` (`locking_regression_test.go`)

All tests pass with `-race`.

## Breaking changes

None. Existing behavior is unchanged when the lock is free.


nsfilho:
image

Thank you so much for taking the initiative to try and solve this issue! However, it seems there might still be a bug in the code. After running wacli --follow, when I try to send a message, it just hangs indefinitely.

I hope the attached image helps shed some light on it. Thanks in advance if you have any ideas about what might be going on!


Final error message:

remote send error: failed to get device list: failed to send usync query: info query timed out

sserrano44:
Tested on macOS with a locally built wacli 0.5.0.

Build:
go build -tags sqlite_fts5 -o ./dist/wacli ./cmd/wacli

Repro before this change:
- Terminal 1: ./dist/wacli sync --follow
- Terminal 2: ./dist/wacli send text --to "12036340724.....@g.us" --message "Test"
- Previous behavior: store is locked (another wacli is running?)

Result with current build:
- Same two-terminal flow now works
- Message sent successfully while sync --follow remained running
- Output:
  Sent to 1203634072....@g.us (id ...)

Notes:
- I used ./dist/wacli explicitly in both terminals to avoid mixing with any older brew-installed binary.
- In my case this fixes the original lock-contention workflow from #70.
- I did not hit the hanging/timeout issue another commenter mentioned.

Umair444:
I have tested it on Arch and this patch is working. @steipete, please merge it into main.

steipete:
Thanks @najibninaba. I rebuilt the send-while-sync path on current `main` in eaa7a1b with a smaller scoped implementation: follow-mode sync hosts a private store-local `.send.sock`, and `send text/file/voice/react` delegate to it only when the store lock is held.

I kept this out of the older broad refactor because main has moved a lot since this PR. Proof: local full gate passed, main CI is green, and live verification succeeded with `sync --follow` holding the lock while a separate `send text --json` delegated successfully.

Closing this PR as superseded by the main implementation. Appreciate the detailed design and tests here; it helped validate the shape.

PR #95, opened 2026-03-20
feat: automatic link preview for text messages

### LINKED ISSUE #94: feat: automatic link preview when sending text messages
When sending a text message containing a URL, wacli could automatically fetch Open Graph metadata (title, description, thumbnail) and attach it as a WhatsApp link preview.

**Proposed behavior:**
- `wacli send text --to ... --message "Check this out https://example.com"` → sends with link preview
- `--no-preview` flag to disable it
- 10s timeout for fetching metadata, graceful fallback to plain text on failure

I have a working implementation ready if this is welcome.

elgatoflaco:
## Summary
- Automatically fetch Open Graph metadata (title, description, thumbnail) when sending a text message containing a URL
- New `internal/linkpreview` package handles URL detection, OG parsing, and thumbnail fetching
- New `SendTextWithPreview` method on the WA client
- `--no-preview` flag to opt out
- Graceful fallback to plain text on fetch failure (10s timeout)

Closes #94

## Test plan
- [ ] Send a message with a URL → verify link preview appears in WhatsApp
- [ ] Send with `--no-preview` → verify no preview
- [ ] Send a message without URLs → verify normal behavior
- [ ] Test with unreachable URL → verify graceful fallback


steipete:
Thanks @elgatoflaco. I rebuilt this on current `main` rather than merging the stale branch, because the send path changed quite a bit since this PR: recipient resolution now supports contacts/groups/chats with `--pick`, text sends can quote replies, and sends keep the connection alive briefly for retry receipts.

Landed on `main` as `ed4df0b`.

What shipped:
- `wacli send text` now fetches Open Graph metadata for the first `http://` or `https://` URL and sends it as a WhatsApp `ExtendedTextMessage` link preview.
- Added `--no-preview` to opt out.
- Preview fetches are bounded to 10s and fail closed to plain text before sending, so we do not risk duplicate sends by trying preview-send then fallback-send after an uncertain WhatsApp error.
- Kept reply/send retry behavior from current main intact.
- Used `golang.org/x/net/html`, which is already in the dependency graph via whatsmeow, instead of regex-parsing HTML.

Proof:
- local full gate: `pnpm format:check && pnpm lint && pnpm test && pnpm build && git diff --check`
- main CI for `ed4df0b`: https://github.com/steipete/wacli/actions/runs/25352360922

Closing this PR as superseded by the maintainer rebuild on main. Thanks again for proposing the feature and the initial implementation.

PR #97, opened 2026-03-21
feat: add read-only mode for safe AI agent access

### LINKED ISSUE #96: Feature: Read-only mode (--readonly / WACLI_READONLY)
## Use Case

AI agents and automation tools often need access to WhatsApp history (search, sync, backfill) but should **not** be able to send messages or mutate groups. Currently there's no built-in way to enforce this — it relies on the caller's discipline.

## Proposal

Add a `--readonly` global flag and `WACLI_READONLY=1` environment variable that disables all write operations:

- `send text` / `send file`
- `groups rename` / `leave` / `join`
- `groups invite revoke`
- `groups participants add/remove/promote/demote`

When enabled, any write command returns a clear error:
```
wacli is running in read-only mode (--readonly or WACLI_READONLY=1); write operations are disabled
```

Read operations (chats list, messages search, history backfill, sync, etc.) continue to work normally.

## Why env var?

The env var makes it easy to enforce read-only at the deployment level (e.g., in a container or agent config) without modifying every command invocation.

## Implementation

I have a working implementation ready — happy to open a PR. It's ~38 lines across 6 files, all guarding the `RunE` of write commands.

otmanesabir:
Closes #96

## Problem

AI agents and automation tools need WhatsApp history access (search, sync, backfill) but should **never** be able to send messages or mutate groups.

## Solution

Add a `--readonly` flag to `wacli auth` that permanently disables all write operations for the linked device. The setting is persisted in `wacli.db` (new `settings` table, migration v4).

### Usage

```bash
wacli auth --readonly
```

Once set, all write commands are blocked until re-authenticated without `--readonly`.

### What's blocked
- `send text` / `send file`
- `groups rename` / `leave` / `join`
- `groups invite revoke`
- `groups participants add/remove/promote/demote`

### What still works
- `chats list`, `messages search`, `history backfill`, `sync`
- `contacts`, `groups list/info`, `media download`
- `auth status` (shows readonly state)

## Implementation
- New `settings` key-value table in `wacli.db` (migration v4)
- Write commands tagged with `Annotations{"mutates": "true"}`
- Single check in root `PersistentPreRunE` — no scattered if-statements
- `auth status` reports readonly in both human and JSON output

One squashed commit, 134 lines added across 9 files.

### DIFF
--- a/cmd/wacli/auth.go
+++ b/cmd/wacli/auth.go
@@ -18,6 +18,7 @@ func newAuthCmd(flags *rootFlags) *cobra.Command {
 	var follow bool
 	var idleExit time.Duration
 	var downloadMedia bool
+	var readOnly bool
 
 	cmd := &cobra.Command{
 		Use:   "auth",
@@ -31,6 +32,11 @@ func newAuthCmd(flags *rootFlags) *cobra.Command {
 				return err
 			}
 			defer closeApp(a, lk)
+
+			// Persist readonly setting if requested (saved to wacli.db after sync)
+			if readOnly {
+				flags.readOnly = true
+			}
 
 			mode := appPkg.SyncModeBootstrap
 			if follow {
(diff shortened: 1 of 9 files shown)


steipete:
Thanks @otmanesabir. The stateless read-only mode landed on main in f6a9454 as --read-only plus WACLI_READONLY. I kept it invocation/env based rather than persisting the setting in the store, so closing this variant as superseded.

Issue #91, opened 2026-03-09
wacli send command consistently fails with "info query timed out" after authentication


fasterthinking:
Subject: wacli send command consistently fails with "info query timed out" after authentication

Description:
wacli version 0.5.0, running on Linux, consistently fails to send messages with the error failed to get device list: failed to send usync query: info query timed out. This occurs even after successful authentication, re-linking, a fresh wacli build, and ensuring the linked device's WhatsApp app is active/foregrounded.

Steps to Reproduce:

1. Install wacli (e.g., go install github.com/steipete/wacli/cmd/wacli@main).
2. Run wacli auth and successfully link a device by scanning the QR code with WhatsApp.
3. Attempt to send any message: /root/go/bin/wacli send text --to "<phone_number>" --message "Test message".

Expected Result:
Message is sent successfully via WhatsApp.

Actual Result:
After approximately 30-60 seconds, the command exits with the error:
failed to get device list: failed to send usync query: info query timed out

The wacli doctor command correctly reports AUTHENTICATED: true, but CONNECTED: false. Attempts to keep wacli sync --follow running in the background (via systemd user service) also result in the same timeout when wacli send is attempted. Stopping the sync service, running wacli sync --once, and immediately attempting wacli send also fails.

Environment:

• Operating System: Linux (Ubuntu-based)
• Go Version: go1.22.2 (though wacli automatically switches to go1.25.8 during installation)
• wacli Version: 0.5.0
• WhatsApp App: Latest version on iOS/Android (confirmed foregrounded/unlocked during send attempts)

Androz2091:
Hi, I'm getting the same issue

Androz2091:
I can reproduce the issue with MacOS 0.2.0

steipete:
Fixed/mitigated on main in 91d240a. Send now retries once after reconnect for info-query/usync timeout failures and returns a bounded error instead of hanging indefinitely.

PR #98, opened 2026-03-22
fix: normalize phone numbers with leading plus

LevyDeSales:
## Summary
- strip a leading `+` from phone-number inputs before building a user JID
- preserve existing behavior for explicit JIDs containing `@`
- add a regression test covering whitespace + `+` normalization

## Why
Several reports show `wacli send text --to "+49123456789" ...` timing out while `49123456789` works.
The malformed lookup path builds `+49123456789@s.whatsapp.net`, which leads to `info query timed out` / user-info lookup failures.

This is a small, low-risk fix that addresses the phone-number parsing bug independently of the larger IPC/send-delegation work in #92.

## Testing

## Related
- Fixes/addresses #28
- Likely improves symptoms reported in #9 and #91
- Separate from #92 (IPC delegation while `sync --follow` is running)

### DIFF
--- a/internal/wa/client.go
+++ b/internal/wa/client.go
@@ -255,6 +255,7 @@ func ParseUserOrJID(s string) (types.JID, error) {
 	if strings.Contains(s, "@") {
 		return types.ParseJID(s)
 	}
+	s = strings.TrimPrefix(s, "+")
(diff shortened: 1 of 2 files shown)


LevyDeSales:
Closing this because the work will continue in the maintainer's fork instead of upstream.

PR #102, opened 2026-03-28
feat: add multi-account support

elgatoflaco:
## Summary

- Add `--account` / `-a` global flag and `WACLI_ACCOUNT` env var to select which WhatsApp account to use
- Each account stored in `~/.wacli/accounts/<name>/` with isolated session, database, lock, and media
- New `wacli accounts` command: `list`, `default [name]`, `remove <name> --force`
- Auto-migrate legacy single-account layout to `accounts/default/` on first run
- Account name validation to prevent path traversal
- Per-account file locking enables concurrent use of different accounts

## Usage

```bash
# Auth a second account
wacli -a work auth

# All commands work with -a
wacli -a work messages search "hello"

# Manage accounts
wacli accounts list
wacli accounts default work
```

## Backward compatibility

Existing users are seamlessly migrated: on first run, if `session.db` is found in the root store dir, all files are moved to `accounts/default/`. Everything works as before without `-a`.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [x] `go vet ./...` passes
- [x] Unit tests for account name validation, resolution, listing, and legacy migration
- [x] Manual testing: authenticated two accounts on a separate machine via `-a`

### DIFF
--- a/cmd/wacli/accounts.go
+++ b/cmd/wacli/accounts.go
@@ -1,0 +1,172 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"text/tabwriter"
+
+	"github.com/spf13/cobra"
+	"github.com/steipete/wacli/internal/config"
+	"github.com/steipete/wacli/internal/out"
+)
+
+func newAccountsCmd(flags *rootFlags) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "accounts",
+		Aliases: []string{"account"},
+		Short:   "Manage WhatsApp accounts",
+	}
+
+	cmd.AddCommand(newAccountsListCmd(flags))
+	cmd.AddCommand(newAccountsDefaultCmd(flags))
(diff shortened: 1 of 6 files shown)


Abacha234 (approved):

steipete:
Thanks for putting this together and for testing it with two accounts. I am closing it for now rather than trying to rebase it.

Current `main` already has the safer primitive for multiple independent sessions: `--store DIR` and `WACLI_STORE_DIR` select an entirely separate store directory, with its own `session.db`, `wacli.db`, media directory, and lock. That covers the operational need without silently migrating every existing user from `~/.wacli` into a new nested `accounts/default/` layout.

The remaining proposal is a larger product decision: account registry, default account state, removal semantics, migration, docs, and compatibility for scripts that already pass `--store`. That is too much to land from a stale/conflicting branch without a current issue and live migration proof.

If we revisit this, I would start smaller on current `main`: document the separate-store pattern first, then add an `accounts` convenience layer only if users still need it and we can prove the migration path end-to-end.

PR #105, opened 2026-04-05
fix(doctor): surface lock-owned paired sessions

artemgetmann:
## Summary

`wacli doctor` currently reports `authenticated=true` and `connected=false` when another long-lived `wacli sync --follow` process already owns the store lock.

That makes the paired-but-owned state look like a plain disconnect, even though the real situation is “this process cannot verify connectivity because another process owns the session.”

This PR keeps `connected` backward-compatible and adds explicit lock-owner reporting:

- `lock_owner_pid`
- `connection_state`

When the store is paired and locked by another process, `connection_state` now reports `locked_by_other_process` instead of silently looking like an ordinary disconnected state.

## Why

A long-lived sync owner is the normal healthy shape for `wacli` live sync. In that state, `doctor` should at least surface that another process owns the paired session instead of only emitting `connected=false`.

## Validation

- `go test ./cmd/wacli/...`

### DIFF
--- a/cmd/wacli/doctor.go
+++ b/cmd/wacli/doctor.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"text/tabwriter"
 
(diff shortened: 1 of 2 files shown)


steipete:

steipete:
Closing as already landed on main in 0ce9208 (#105). Thanks @artemgetmann.

PR #109, opened 2026-04-10
store: persist archive backfill coverage state

cropsgg:
Add a dedicated history backfill state table and coverage queries so archive progress can be resumed and inspected without relying on ad hoc scripts. This gives later backfill orchestration a stable, test-covered persistence layer for blocked, stalled, and completed chat state.

### DIFF
--- a/internal/store/history.go
+++ b/internal/store/history.go
@@ -1,0 +1,348 @@
+package store
+
+import (
+	"database/sql"
+	"fmt"
+	"strings"
+	"time"
+)
+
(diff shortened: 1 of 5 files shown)


steipete:
Thanks for splitting this out. I’m closing this partial backfill slice as superseded by #111, which contains this work plus the engine/CLI layer in one reviewable proposal. We’ll keep #111 as the canonical thread for the resumable multi-chat/archive-coverage design.

PR #111, opened 2026-04-10
Feature/history backfill cli

cropsgg:
Expose archive coverage inspection and multi-chat fill commands so users can drive resumable history recovery from the CLI instead of shell loops. This gives the new backfill engine a clear human and JSON interface while preserving the existing single-chat backfill escape hatch.

### DIFF
--- a/cmd/wacli/history.go
+++ b/cmd/wacli/history.go
@@ -5,20 +5,186 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
+	"text/tabwriter"
 	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/steipete/wacli/internal/app"
 	"github.com/steipete/wacli/internal/out"
+	"github.com/steipete/wacli/internal/store"
 )
 
 func newHistoryCmd(flags *rootFlags) *cobra.Command {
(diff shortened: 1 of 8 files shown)


steipete:
Thanks @cropsgg — I rebuilt the safe/current slice of this on main as a2c7803.\n\nWhat landed:\n- `history coverage` to inspect local archive anchors from the current `wacli.db` schema\n- `history fill --dry-run` to plan matching ready chats without connecting to WhatsApp or writing state\n- README + docs/history + quickstart/spec/index coverage\n- regression tests for the store coverage query and CLI dry-run behavior\n\nI did not land the stale branch directly because it now conflicts with current backfill internals and predates the newer OOM-safe/manual on-demand history-sync download path plus NDJSON event behavior. The persisted resume table / live multi-chat fill can be added later as a separate migration-backed step.\n\nVerified:\n- checked pinned whatsmeow history-sync contract: on-demand requests return `ON_DEMAND` history, recommended count is 50, and manual downloads are only needed when manual history sync download is enabled\n- live read-only smoke on the real store: `history coverage`, `history fill --dry-run`, JSON output\n- full local gate: `pnpm format:check && pnpm lint && pnpm test && pnpm build && git diff --check`\n\nClosing this PR as superseded by the maintainer rewrite on main.

PR #114, opened 2026-04-10
feat: add NDJSON lifecycle events

0xatrilla:
## Summary

Add a new global `--events` flag that emits newline-delimited JSON lifecycle events to `stderr` for long-running flows while preserving the existing human-readable output when the flag is not used.

This wires event emission into:
- `wacli auth`
- `wacli sync`
- `wacli history backfill`

## Events

Current event types include:
- `auth_starting`
- `qr_code`
- `connected`
- `disconnected`
- `history_sync`
- `progress`
- `reconnecting`
- `stopping`
- `idle_exit`
- `backfill_requesting`
- `backfill_response`
- `backfill_stopped`

## Implementation

- Add `internal/out/EventWriter` for thread-safe NDJSON event emission
- Add persistent `--events` flag on the root command
- Thread an optional event writer through `app.Options`
- Emit structured lifecycle events alongside existing stderr output in auth/sync/backfill paths
- Add focused tests for the event writer and lifecycle emission

## Testing

- `go test ./...`
- `go run ./cmd/wacli --help`

## Notes

Normal CLI behavior is unchanged unless `--events` is passed.

### DIFF
--- a/cmd/wacli/auth.go
+++ b/cmd/wacli/auth.go
@@ -38,6 +38,7 @@ func newAuthCmd(flags *rootFlags) *cobra.Command {
 			}
 
 			fmt.Fprintln(os.Stderr, "Starting authentication…")
+			_ = a.Events().Emit("auth_starting", nil)
 			res, err := a.Sync(ctx, appPkg.SyncOptions{
 				Mode:            mode,
 				AllowQR:         true,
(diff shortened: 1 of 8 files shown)


dinakars777 (commented):
Thanks for putting this together. I like the direction, and this is the cleaner remaining PR for machine-readable lifecycle events now that the older events PR is closed.

I can’t merge this as-is though. It currently conflicts with current `main`, and there is a correctness issue with the NDJSON stream: `--events` writes JSON to `stderr` while existing human progress output also writes carriage-return status text to `stderr`, which can produce mixed/non-JSON lines and break line-by-line parsers.

Could you please rebase on current `main` and adjust the implementation so `--events` guarantees valid NDJSON output, ideally with a regression test that proves event output remains parseable during sync progress? Once that is fixed, this should be a much cleaner review target.

dinakars777:
Update: I went ahead and rebuilt this on current `main` in #204 so we can land the feature with the NDJSON correctness fix and regression coverage in place.

Thanks again for the original implementation and direction here. Closing this as superseded by #204.

Issue #104, opened 2026-04-04
send text hangs with no output after successful auth; sync can connect briefly but doctor shows CONNECTED false


sls74d-creator:
I’m seeing a reproducible send failure on macOS with `wacli 0.2.0` installed via Homebrew.

## Environment
- macOS (Apple Silicon)
- `wacli 0.2.0`
- installed via `brew install steipete/tap/wacli`

## Observed behavior
- `wacli auth` succeeds
- `wacli doctor` shows:
  - `AUTHENTICATED true`
  - `CONNECTED false`
- `wacli sync --once` does connect briefly and exits cleanly
- after that, `wacli send text --to "+447900424242" --message "WhatsApp test from Benson: message path is live."` hangs with no output
- `wacli --json send text ...` also hangs with no output
- if `sync` is running, `send` fails immediately due to store lock, which is expected
- when not locked, `send` still hangs indefinitely

## Related errors seen during auth/sync
- `Failed to do initial fetch of app state regular_high: failed to decode app state regular_high patches: failed to verify patch v12: mismatching LTHash`

## Recovery attempts
- `wacli auth logout`
- full local store reset by moving `~/.wacli`
- fresh re-auth
- reinstall stable via Homebrew
- tried installing `--HEAD`, but current HEAD failed to build from the Homebrew tap

## Result
- auth can be established
- one-shot sync can connect
- sending never completes and produces no output

## Question
Is this a known issue around app-state/session handling or send path deadlocking after partial connection state?

pedro-correia-tech:
Same behaviour on my side

sls74d-creator:
Additional data point from testing on macOS stable (0.2.0):

- auth succeeds
- `doctor` often ends up at `AUTHENTICATED=true`, `CONNECTED=false`
- `sync --once` can connect briefly and exit cleanly
- `send text` then hangs with no output when sync is not running
- if sync *is* running, send fails immediately on the expected store lock

There are at least a few related reports that look like the same family of send/connectivity problems:
- #103 (`send` timeout / LID cache path)
- #91 (send timeout after authentication)
- #70 (lock contention while sync is running)

So this may not be an isolated auth problem so much as a broader send-path issue after partial/unstable session connectivity on macOS.

Pedro also reports the same behavior here, so this is at least reproducible across more than one environment.

sls74d-creator:
Fresh stronger repro data point:

I repeated the test with a completely fresh store and a different WhatsApp account (not the original/self-send scenario):

[...]

Issue #44, opened 2026-02-16
send text fails with "failed to get user info to fill LID cache: info query timed out"


bauer-jan:
All wacli send commands fail with a LID cache timeout, even after a fresh auth and full sync. Read-only commands work fine.

Environment

wacli 0.2.0 (Homebrew)
macOS arm
Fresh install + auth
Steps to reproduce

rm -rf ~/.wacli && wacli auth (scan QR, wait for full sync)
Sync completes: "Authenticated. Messages stored: 8452"
wacli doctor shows AUTHENTICATED=true
wacli chats list works fine, shows conversations
wacli send text --to +49XXXXXXXXXX --message "Hi" → hangs, then fails
Error

failed to get user info for +49XXXXXXXXXX@s.whatsapp.net to fill LID cache: failed to send usync query: info query timed out
What I tried

Sending to multiple different numbers (both in synced contacts and not)
Fresh rm -rf ~/.wacli + wacli auth with full sync (8452 messages, 335 conversations)
Phone open on WhatsApp during send attempts
wacli --timeout 120s send text ...
All read commands work: chats list, messages search, contacts search
Expected behavior wacli send text should send the message after a successful auth and sync.

bauer-jan:
I just saw this is a duplicate to https://github.com/steipete/wacli/issues/9, closing the issue

steipete:
Fixed/mitigated on main in 91d240a. Send now treats info-query/usync timeouts as retryable once after reconnect, with a bounded send attempt timeout.

PR #115, opened 2026-04-10
fix: bound send hangs and retry stale sessions

0xatrilla:
## Summary

Fix the send-path failure mode where `wacli send text` / `wacli send file` can appear to hang indefinitely or fail after a long wait when the session is authenticated but the underlying WhatsApp connection is stale or unhealthy.

This targets the issue cluster around:
- #104
- #91
- #44

## What changed

- Add a shared send helper that runs each send attempt with a bounded per-attempt timeout
- Return a clear error like `send timed out after 45s` instead of silently hanging
- Retry once after a clean reconnect when the failure looks like a stale/disconnected session or a usync/LID lookup timeout
- Apply the same bounded/retrying behavior to both `send text` and `send file`

## Retry conditions

The reconnect + retry path only kicks in for clearly connection-shaped failures, including:
- `info query timed out`
- `failed to send usync query`
- `failed to get user info`
- `failed to get device list`
- `not connected`

This keeps ordinary validation or file errors unchanged.

## Testing

- `go test ./cmd/wacli`
- `go test ./...`

## Notes

This does not try to redesign the full send/LID flow. The goal here is to make the current send path much safer operationally by avoiding indefinite stalls and recovering once from stale-session failures.

### DIFF
--- a/cmd/wacli/send.go
+++ b/cmd/wacli/send.go
@@ -55,7 +55,12 @@ func newSendTextCmd(flags *rootFlags) *cobra.Command {
 				return err
 			}
 
-			msgID, err := a.WA().SendText(ctx, toJID, message)
+			msgID, err := runSendOperation(ctx, func(ctx context.Context) error {
+				a.WA().Close()
+				return a.Connect(ctx, false, nil)
+			}, func(ctx context.Context) (string, error) {
+				return a.WA().SendText(ctx, toJID, message)
+			})
 			if err != nil {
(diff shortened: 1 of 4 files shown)


steipete:
Thanks @0xatrilla. Reworked and landed this on main in 91d240a: send text/file/react attempts are bounded, retry once after reconnect for stale-session/usync timeout errors, and have tests. Closing as superseded.

PR #116, opened 2026-04-10
fix: normalize phone-style recipient input

fahmidme:
## Summary
Normalize phone-style recipient input so common human-entered values work without requiring exact JIDs.

### Changes
- accept `+E164` input like `+15551234567`
- accept formatted phone input like `+1 (555) 123-4567`
- preserve full JIDs like `1234567890@s.whatsapp.net` and group JIDs like `123@g.us`
- return a clearer error for invalid phone-style input containing unsupported characters
- clarify accepted recipient formats in CLI help text and README

### Testing
- manual verification after rebuilding the binary

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ pnpm wacli history backfill --chat 1234567890@s.whatsapp.net --requests 10 --cou
 
 # Send a message
 pnpm wacli send text --to 1234567890 --message "hello"
+# phone numbers can also be passed as +E164 or formatted input like "+1 (234) 567-8900"
 
 # Send a file
 ./wacli send file --to 1234567890 --file ./pic.jpg --caption "hi"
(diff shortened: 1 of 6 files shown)


ImLukeF:
Review note: `ParseUserOrJID` is currently too permissive for phone-style input. It accepts `+` anywhere in the string and treats Unicode digits as valid, so malformed input like `12+34` or `١٢٣٤` can be normalized into the wrong recipient JID instead of being rejected. I verified a tighter parser locally by allowing only a leading `+` and ASCII digits, and added regression coverage for both cases. I wasn’t able to push that patch back because GitHub rejected branch updates from my account with `403 Permission denied`.

steipete:
Thanks @fahmidme. I am closing this version because the parser shape here is too permissive for recipient input: it accepts plus signs away from the first character and Unicode digits. Main now has stricter recipient validation with coverage for plus placement, ASCII digits, length, and invalid punctuation. The remaining question of whether to accept formatted phone numbers should continue in #130 or a fresh narrower PR.

PR #117, opened 2026-04-10
fix: update whatsmeow to fix 405 client outdated error

alitala2009:
## Problem

`wacli` was failing to connect to WhatsApp with a **405 (Client Outdated)** error:

```
Client outdated (405) connect failure (client version: 2.3000.1030403648)
```

## Root Cause

The pinned `go.mau.fi/whatsmeow` dependency (commit `20260211`) advertises an old client version string (`2.3000.1030403648`) that WhatsApp's servers now reject, preventing any connection from being established.

## Fix

Bump `go.mau.fi/whatsmeow` to the latest commit (`20260410`), which includes an updated client version string that passes WhatsApp's current version check. Also bumps the transitive dependency `go.mau.fi/util` from v0.9.5 to v0.9.6.

```
go.mau.fi/whatsmeow v0.0.0-20260211193157-7b33f6289f98 => v0.0.0-20260410162419-b95d92207080
go.mau.fi/util v0.9.5 => v0.9.6
```

## Testing

Built successfully with `GOOS=linux GOARCH=amd64 go build ./cmd/wacli/`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/mattn/go-sqlite3 v1.14.34
 	github.com/mdp/qrterminal/v3 v3.2.1
 	github.com/spf13/cobra v1.10.2
-	go.mau.fi/whatsmeow v0.0.0-20260211193157-7b33f6289f98
+	go.mau.fi/whatsmeow v0.0.0-20260410162419-b95d92207080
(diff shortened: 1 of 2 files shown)


dinakars777:
Merged, Thanks @alitala2009.

PR #118, opened 2026-04-10
docs: refresh README with architecture diagram

fahmidme:
## Summary
Refresh the README with a clearer project narrative, cleaner quick-start sections, and an architecture diagram checked into the repo.

### What changed
- add a repo-local architecture image under `docs/images/architecture-light.png`
- rewrite the README intro around what `wacli` actually does
- add a dedicated architecture section explaining the CLI, app layer, local store, locking, FTS, and WhatsApp transport path
- clean up install, quick start, common workflows, storage model, and JSON output sections
- keep claims aligned with the current `main` branch behavior

### Why
The existing README had the right ingredients, but it felt more like raw project notes than a polished open-source front door. This refresh makes the repo easier to understand at a glance and gives the project a much stronger visual and conceptual entry point.

### Notes
- the image is an optimized repo-local copy derived from the selected architecture render
- this PR is docs-only


steipete:
Thanks for the README pass and the architecture diagram. I rechecked this against current `main` today and I am going to close it as stale rather than land it as-is.

Current `README.md` has moved quite a bit since this branch was opened. Applying this now would drop newer documented behavior, including:
- Linux XDG store-dir default plus legacy `~/.wacli` fallback
- cgo / `go-sqlite3` build requirements and Homebrew cgo rebuild note
- quoted text/file replies, reactions, presence, `--full`, `--read-only`, `--lock-wait`
- the expanded current command surface and safety/diagnostics notes

The PR also adds a large binary architecture image, which is not worth taking while the text itself is behind the current CLI surface. The useful narrative bits should be revived in a fresh README edit against current `main`, ideally without regressing the command/docs details that have since landed.

PR #119, opened 2026-04-10
fix: sync --once hangs indefinitely when non-message events keep resetting idle timer

### LINKED ISSUE #87: bug: sync --once hangs indefinitely at "Connected." when no history sync event is received
## Bug Report

**wacli version:** 0.2.0
**OS:** Linux (Docker container, x86_64)
**Install method:** binary

## Description

When running `wacli sync --once`, the process connects successfully and prints `Connected.` but then hangs indefinitely without ever exiting, even after the `--idle-exit` timeout (default 30s) has elapsed.

## Steps to Reproduce

```bash
wacli sync --once
# or
wacli sync --once --idle-exit 10s
```

## Expected Behavior

After connecting and being idle for `--idle-exit` duration (30s by default), the process should exit automatically.

```
Connected.
Idle for 30s, exiting.
```

## Actual Behavior

```
Connected.
```

…and the process hangs indefinitely. It never exits, even with `--idle-exit` set to a short duration.

## Root Cause (suspected)

This appears to happen on **subsequent syncs** (after the initial `wacli auth`). WhatsApp only pushes a full `HistorySync` event on the first connection. On subsequent connections, if no new messages arrive, no events are dispatched at all.

The idle timer in `--once` mode likely only starts after at least one event (e.g. `HistorySync`) is received. If the account is already fully synced and no new messages arrive, the idle timer never starts → process hangs forever at `Connected.`

This is distinct from `--follow` mode (which is meant to run indefinitely) and specifically affects the `--once` use case, which is the correct flag for periodic/scripted syncs.

## Workaround

```bash
# Force exit via timeout wrapper
timeout 60 wacli sync --once || true
```

## Related

- #61 — sync cleanup can hang after SIGINT (similar: process fails to exit cleanly)
- #85 (PR) — adds `--events` NDJSON lifecycle output which would help diagnose this

## Additional Context

The `--idle-exit` flag description reads "exit after being idle (once mode)" — this implies the timer should fire even when no events are received, i.e., immediately after connecting if the store is already up to date. If that is the intended design, the timer is not being started correctly in the no-event-received path.

jyothepro:
## Summary

Fixes #87 — `sync --once` hangs indefinitely at "Connected." when no history sync event is received.

## Root Cause

The idle timer in the sync event handler was being reset on **every** whatsmeow event — not just messages and history syncs. This includes `Connected`, `Disconnected`, presence updates, read receipts, keep-alive pings, and any other event type that whatsmeow emits.

```go
// BEFORE (broken): resets on ALL events
handlerID := a.wa.AddEventHandler(func(evt interface{}) {
    lastEvent.Store(time.Now().UTC().UnixNano()) // ← fires for every event type
    switch v := evt.(type) {
    case *events.Message: ...
    case *events.HistorySync: ...
    case *events.Connected: ...      // timer already reset above
    case *events.Disconnected: ...   // timer already reset above
    }
})
```

Since WhatsApp regularly sends non-message events (presence, receipts, etc.), the idle timer never reaches its threshold and `sync --once` never exits — it hangs until the user kills the process.

## Fix

Moved `lastEvent.Store(...)` from the top of the handler (before the switch) into only the two cases that should actually keep the sync alive:

- `*events.Message` — a live incoming message arrived
- `*events.HistorySync` — a history sync batch is being processed

All other events (`Connected`, `Disconnected`, presence, receipts, etc.) no longer reset the idle timer.

```go
// AFTER (fixed): resets only on message/history events
handlerID := a.wa.AddEventHandler(func(evt interface{}) {
    switch v := evt.(type) {
    case *events.Message:
        lastEvent.Store(time.Now().UTC().UnixNano()) // ← only here
        ...
    case *events.HistorySync:
        lastEvent.Store(time.Now().UTC().UnixNano()) // ← and here
        ...
    }
})
```

## Regression Test

Added `TestSyncOnceExitsWithNonMessageEvents` which:
1. Starts `sync --once` with a 200ms idle timeout
2. Continuously emits `Connected` events every 30ms during the idle wait
3. Asserts sync exits within 1.5s

**Without the fix**, this test **fails** — sync hangs for the full 5s context timeout:
```
sync_test.go:271: sync --once took 5.000332167s; non-message events should not reset the idle timer
--- FAIL: TestSyncOnceExitsWithNonMessageEvents (5.01s)
```

**With the fix**, this test **passes** — sync exits in ~260ms as expected:
```
Idle for 200ms, exiting.
--- PASS: TestSyncOnceExitsWithNonMessageEvents (0.26s)
```

## Test plan

- [x] `TestSyncOnceExitsWithNonMessageEvents` — proves non-message events don't reset idle timer
- [x] `TestSyncOnceIdleExit` — existing test still passes (idle exit with no events)
- [x] `TestSyncStoresLiveAndHistoryMessages` — existing test still passes (messages still stored correctly)
- [x] Full test suite passes: `go test ./...`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Send: strip a leading `+` from phone-number recipients before building WhatsApp JIDs. (#74 — thanks @FrederickStempfle)
 - Search: keep FTS5 enabled after reopening existing databases with already-applied migrations. (#185 — thanks @iamhitarth)
 - Send: persist retry-message plaintext so linked devices can decrypt retried messages. (#186 — thanks @SimDamDev)
+- Sync: keep `sync --once` idle timing focused on message/history events so connection chatter cannot hang exit. (#119 — thanks @jyothepro)
 - Sync: start `sync --once` idle timing after the `Connected` event. (#171 — thanks @fuleinist)
 - Sync: include event type, stack trace, and recovery count when logging recovered event-handler panics. (#181 — thanks @shaun0927)
 - Windows: split store locking by platform so the lock package compiles on Windows. (#188 — thanks @dinakars777)
--- a/internal/app/sync.go
+++ b/internal/app/sync.go
@@ -69,6 +69,7 @@ func (a *App) Sync(ctx context.Context, opts SyncOptions) (SyncResult, error) {
 	if err := a.Connect(ctx, opts.AllowQR, opts.OnQRCode); err != nil {
 		return SyncResult{}, err
 	}
+	lastEvent.Store(time.Now().UTC().UnixNano())
 
 	if opts.DownloadMedia {
 		var err error
--- a/internal/app/sync_events.go
+++ b/internal/app/sync_events.go
@@ -44,12 +44,12 @@ func (a *App) addSyncEventHandler(ctx context.Context, opts SyncOptions, message
 					n, evt, r, debug.Stack())
 			}
 		}()
-		lastEvent.Store(time.Now().UTC().UnixNano())
-
 		switch v := evt.(type) {
 		case *events.Message:
+			lastEvent.Store(time.Now().UTC().UnixNano())
 			a.handleLiveSyncMessage(ctx, opts, v, messagesStored, enqueueMedia)
 		case *events.HistorySync:
+			lastEvent.Store(time.Now().UTC().UnixNano())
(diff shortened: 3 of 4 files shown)


steipete:

- Land commit: 894bc5d

Thanks @jyothepro!

PR #120, opened 2026-04-11
fix: sanitize user input in FTS5 search queries

### LINKED ISSUE #57: bug: FTS5 query syntax exposed to user input without sanitization
## Description

The user's search query is passed directly to the FTS5 `MATCH` clause (`internal/store/search.go:47-55`). FTS5 has its own query syntax (AND, OR, NOT, NEAR, column filters, prefix `*`). Users can craft unexpected queries or cause expensive operations.

**Severity:** Low

**File:** `internal/store/search.go:47-55`

## Suggested Fix

Wrap user queries in double-quotes to treat as phrase literals, escaping internal double-quotes. Fix submitted in PR from `alexander-morris/wacli:security/audit-fixes`.

jyothepro:
## Summary

Fixes #57 — FTS5 query syntax is exposed to user input without sanitization, causing crashes and unexpected search behavior.

## Root Cause

`searchFTS()` passes raw user input directly as an FTS5 `MATCH` expression:

```go
// BEFORE (vulnerable)
WHERE messages_fts MATCH ?
// args = []interface{}{p.Query}  ← raw user input
```

FTS5 has its own query language with operators (`OR`, `AND`, `NOT`), prefix wildcards (`*`), column filters (`col:`), phrase matching (`"..."`), and proximity searches (`NEAR(...)`). User input containing any of these triggers FTS5 query parsing, not literal text matching.

**Without the fix, these user queries crash:**

| Query | Error |
|---|---|
| `"unclosed` | `unterminated string` |
| `NOT meeting` | `fts5: syntax error near "NOT"` |
| `col:meeting` | `no such column: col` |

## Fix

Added `sanitizeFTSQuery()` that wraps each whitespace-separated token in double quotes, escaping any internal `"` by doubling them (FTS5's escape convention). This forces FTS5 to treat all input as literal text:

```go
func sanitizeFTSQuery(q string) string {
    tokens := strings.Fields(q)
    if len(tokens) == 0 {
        return `""`
    }
    quoted := make([]string, len(tokens))
    for i, tok := range tokens {
        quoted[i] = `"` + strings.ReplaceAll(tok, `"`, `""`) + `"`
    }
    return strings.Join(quoted, " ")
}
```

Examples:
- `hello world` → `"hello" "world"` (literal search, implicit AND)
- `hello OR world` → `"hello" "OR" "world"` (OR treated as literal word)
- `"unclosed` → `"""unclosed"` (quote escaped, no parse error)
- `col:value` → `"col:value"` (no column filter interpretation)

### `TestFTSSearchWithSpecialCharacters`
Runs 7 previously-dangerous queries against a real FTS5 table and asserts none return errors:

**Without the fix — 3 failures:**
```
SearchMessages("\"unclosed") failed: unterminated string
SearchMessages("NOT meeting") failed: fts5: syntax error near "NOT"
SearchMessages("col:meeting") failed: no such column: col
--- FAIL: TestFTSSearchWithSpecialCharacters
```

**With the fix — all pass:**
```
--- PASS: TestFTSSearchWithSpecialCharacters (0.01s)
```

### `TestSanitizeFTSQuery`
Unit tests for the sanitization function covering operators, quotes, column filters, wildcards, NEAR, and whitespace normalization.

## Test plan

- [x] `TestSanitizeFTSQuery` — unit tests for sanitization logic
- [x] `TestFTSSearchWithSpecialCharacters` — dangerous queries no longer crash
- [x] `TestSearchMessagesUsesFTSWhenEnabled` — existing FTS search still works
- [x] Full test suite passes: `go test ./...` and `go test -tags sqlite_fts5 ./...`

### DIFF
--- a/internal/store/search.go
+++ b/internal/store/search.go
@@ -44,6 +44,18 @@ func (d *DB) searchLIKE(p SearchMessagesParams) ([]Message, error) {
 	return d.scanMessages(query, args...)
 }
 
+func sanitizeFTSQuery(q string) string {
+	tokens := strings.Fields(q)
+	if len(tokens) == 0 {
+		return `""`
+	}
+	quoted := make([]string, len(tokens))
+	for i, tok := range tokens {
+		quoted[i] = `"` + strings.ReplaceAll(tok, `"`, `""`) + `"`
+	}
+	return strings.Join(quoted, " ")
+}
+
 func (d *DB) searchFTS(p SearchMessagesParams) ([]Message, error) {
 	query := `
 		SELECT m.chat_jid, COALESCE(c.name,''), m.msg_id, COALESCE(m.sender_jid,''), m.ts, m.from_me, COALESCE(m.text,''), COALESCE(m.display_text,''), COALESCE(m.media_type,''),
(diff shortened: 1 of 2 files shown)


steipete:
Closing as fixed on main: FTS queries are sanitized before MATCH. Thanks @jyothepro.

Issue #84, opened 2026-02-25
history backfill OOM-killed on ARM64 (Raspberry Pi 8GB) — initial history sync blobs exhaust RAM


NePav:
## Environment
- **Device**: Raspberry Pi 4 (8GB RAM), ARM64
- **OS**: Linux 6.12 (Raspberry Pi OS)
- **wacli**: v0.2.0 (dev), installed via `go install`
- **Go**: 1.26.0 linux/arm64

## Problem
`wacli history backfill` is consistently OOM-killed (SIGKILL) on a Raspberry Pi 4 with 8GB RAM, even with minimal settings (`--count 5 --requests 1`).

Meanwhile, `wacli sync --follow` works fine and runs as a persistent systemd service without issues.

## Root Cause Analysis

The issue is in the underlying **whatsmeow** library. When `backfill` calls `a.Sync()` → `a.Connect()`, WhatsApp pushes initial history sync notifications. The processing path in whatsmeow `message.go` (`DownloadHistorySync`):

```go
// 1. Downloads full blob into memory
data, err = cli.Download(ctx, notif)

// 2. Decompresses entirely into RAM via io.ReadAll
rawData, err := io.ReadAll(reader)

// 3. Unmarshals into protobuf structs
err = proto.Unmarshal(rawData, &historySync)
```

For a single history sync chunk, there are **3 copies in memory**: compressed, decompressed, and parsed protobuf. For accounts with significant chat history, this easily exhausts available RAM.

### Additional factors:
1. `historySyncNotifications` channel is buffered to 32 — multiple chunks can queue up and get processed concurrently
2. `storeParsedMessage` calls `GetGroupInfo` and `GetContact` per message — creates many concurrent network calls for large history dumps
3. No streaming decompression — `io.ReadAll` materializes the entire decompressed blob before parsing

### Why `sync --follow` works but `backfill` does not:
`sync --follow` typically connects after the initial history dump was already processed during `wacli auth`. `backfill` creates a new Sync connection each time, and WhatsApp re-pushes the initial history sync blob on connect.

## Reproduction
```bash
# On a memory-constrained device (Pi 4, small VPS, etc.)
wacli history backfill --chat "<any-jid>" --requests 1 --count 5
# Process gets SIGKILL (OOM)

# Even with memory limits:
GOMEMLIMIT=1500MiB wacli history backfill --chat "<jid>" --requests 1 --count 5
# Same result
```

## Possible Mitigations
[...]

PR #121, opened 2026-04-11
fix: drop overflow media jobs instead of spawning unbounded goroutines

### LINKED ISSUE #51: security: goroutine leak in media job enqueuing during large syncs
## Description

When the `mediaJobs` channel (capacity 512) is full during sync, a new goroutine is spawned per overflow job (`internal/app/sync.go:72-78`). During a large history sync with thousands of media messages, hundreds or thousands of goroutines accumulate, each blocking on channel send.

**Severity:** Medium

**File:** `internal/app/sync.go:72-78`

## Current Behavior

```go
go func() {
    select {
    case mediaJobs <- mediaJob{chatJID: chatJID, msgID: msgID}:
    case <-ctx.Done():
    }
}()
```

## Expected Behavior

Drop excess jobs with a log message instead of spawning unbounded goroutines.

## Suggested Fix

Replace the goroutine spawn with a `default` case that logs and drops the job. Fix submitted in PR from `alexander-morris/wacli:security/audit-fixes`.

jyothepro:
## Summary

Fixes #51 — goroutine leak in media job enqueuing during large syncs.

Related: #84 (history backfill OOM on ARM64) — this fix reduces memory pressure during large syncs.

## Root Cause

When the 512-slot media jobs channel is full, the `enqueueMedia` closure spawns a **new goroutine per overflow item** to avoid blocking the event handler:

```go
// BEFORE (leaks goroutines)
default:
    go func() {
        select {
        case mediaJobs <- mediaJob{chatJID: chatJID, msgID: msgID}:
        case <-ctx.Done():
        }
    }()
```

During a large history sync with thousands of media messages, this spawns hundreds of goroutines all blocked on the channel send. With slow or stalled media workers, these goroutines accumulate indefinitely — consuming memory and stack space. On constrained devices (e.g., Raspberry Pi, #84), this amplifies OOM risk.

## Fix

Drop overflow jobs with a log message instead of spawning goroutines. Skipped media can still be downloaded later via `wacli media download`.

```go
// AFTER (bounded, no leak)
default:
    fmt.Fprintf(os.Stderr, "\rmedia queue full, skipping download for %s/%s\n", chatJID, msgID)
```

## Regression Test

`TestSyncMediaEnqueueDropsOverflow` sends 600 media messages (exceeding the 512 channel buffer) with slow download workers (5s delay), then checks goroutine count via `AfterConnect` callback while overflow goroutines would still be blocked.

**Without the fix — FAILS (92 leaked goroutines):**
```
goroutine leak: 92 new goroutines during sync with 600 media messages (before=3, during=95)
--- FAIL: TestSyncMediaEnqueueDropsOverflow (0.08s)
```

**With the fix — PASSES (0 leaked goroutines):**
```
media queue full, skipping download for 123@s.whatsapp.net/m-512
media queue full, skipping download for 123@s.whatsapp.net/m-513
...
--- PASS: TestSyncMediaEnqueueDropsOverflow (0.08s)
```

## Test plan

- [x] `TestSyncMediaEnqueueDropsOverflow` — proves overflow no longer spawns goroutines
- [x] `TestDownloadMediaJobMarksDownloaded` — existing media download test still passes
- [x] `TestSyncStoresLiveAndHistoryMessages` — messages still stored correctly
- [x] Full test suite passes: `go test ./...`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Sync: keep `sync --once` idle timing focused on message/history events so connection chatter cannot hang exit. (#119 — thanks @jyothepro)
 - Sync: start `sync --once` idle timing after the `Connected` event. (#171 — thanks @fuleinist)
 - Sync: include event type, stack trace, and recovery count when logging recovered event-handler panics. (#181 — thanks @shaun0927)
+- Sync: apply bounded backpressure to media download enqueueing instead of spawning unbounded overflow goroutines. (#121 — thanks @jyothepro)
 - Windows: split store locking by platform so the lock package compiles on Windows. (#188 — thanks @dinakars777)
 
 ### Docs
--- a/internal/app/fake_wa_test.go
+++ b/internal/app/fake_wa_test.go
@@ -26,6 +26,7 @@ type fakeWA struct {
 
 	connectEvents []interface{}
 	connectDelay  time.Duration
+	downloadDelay time.Duration
 
 	contacts map[types.JID]types.ContactInfo
 	groups   map[types.JID]*types.GroupInfo
@@ -229,6 +230,13 @@ func (f *fakeWA) DecryptReaction(ctx context.Context, reaction *events.Message) 
 }
 
 func (f *fakeWA) DownloadMediaToFile(ctx context.Context, directPath string, encFileHash, fileHash, mediaKey []byte, fileLength uint64, mediaType, mmsType string, targetPath string) (int64, error) {
+	if f.downloadDelay > 0 {
+		select {
+		case <-time.After(f.downloadDelay):
+		case <-ctx.Done():
+			return 0, ctx.Err()
+		}
+	}
 	if err := os.MkdirAll(filepath.Dir(targetPath), 0o700); err != nil {
 		return 0, err
 	}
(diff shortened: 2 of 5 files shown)


ImLukeF:
Review note: dropping overflow media jobs fixes the goroutine leak, but it also changes `--download-media` semantics by silently losing downloads once the 512-slot queue fills. That is a user-visible regression under large syncs. I validated a local fix that keeps the queue bounded without losing jobs by starting media workers before `Connect` and using bounded backpressure instead of drop-on-overflow, with a regression test that checks both no goroutine leak and full processing of the queued media jobs. I couldn’t push it because GitHub rejected branch updates from my account with `403 Permission denied`.

jyothepro:
@ImLukeF Great catch — you're right that dropping overflow jobs is a user-visible regression. Updated the approach based on your feedback:

**What changed:**
1. Media workers now start **before** `Connect()`, so they drain the channel while events are being emitted
2. The enqueue uses a blocking send (`select { case mediaJobs <- ...: case <-ctx.Done(): }`) instead of drop-on-overflow
3. Workers provide natural backpressure — the event handler pauses briefly when the channel is full, but no jobs are lost

**Result:** No goroutine leak, no dropped downloads. All 600 messages stored in the test, goroutine count stays bounded.

Thanks for the review!

steipete:

- Land commit: 12e8e1a

Thanks @jyothepro!

PR #122, opened 2026-04-11
feat(send): resolve --to against local contacts, groups, and chats

AndroidPoet:
## Summary

`wacli send text`/`send file` only accepted a raw phone number or a full JID, so anyone wanting to message someone had to run `wacli chats list | grep` to copy a JID every time. This PR adds a fuzzy name resolver built on top of the existing local store so that `--to "mom"`, `--to "John Smith"`, or `--to "Family"` just works. JID and phone inputs are unchanged.

```bash
wacli send text --to mom --message "landed"
wacli send text --to "John Smith" --message hi
wacli send text --to Family --message "on my way"
wacli send text --to john --pick 2 --message hi        # scriptable disambiguation
wacli send text --to "+49 170 1234567" --message hi    # phone path unchanged
wacli send text --to 1234567890@s.whatsapp.net ...     # JID path unchanged
```

## Resolution order

1. Input contains `@` → parsed directly as a JID (no DB lookup).
2. **Phone-shaped** input (digits, `+`, spaces, dashes, parens) → take the raw-number path unless the local store holds an *authoritative exact* match (contact name, alias, group name, or group-kind chat row). DM chat rows with numeric display names are ignored so they can't hijack a raw number — this keeps `--to 2024` safe from e.g. a chat titled `Trip 2024`.
3. **Name-shaped** input → ranked fuzzy search of contacts, aliases, groups, and chats: exact > prefix > word-start > substring.

## UX

- One match → send immediately.
- Multiple matches → numbered picker when both `stdin` and `stderr` are terminals (redirecting `stdout` still works; the prompt uses `stderr`).
- `--json` mode, non-interactive sessions, or `--pick N` skip the prompt: either select the Nth match or return a structured "ambiguous" error that lists every candidate, so scripts can parse it.
- Negative `--pick` values are rejected up front; zero stays "unset".
- No match on a name-shaped input → clean error pointing at `wacli contacts search` and the JID escape hatch.
- `EnsureAuthed()` still runs before resolution so unauthenticated users see `run \`wacli auth\`` rather than a misleading "no match".

## Ranking notes

- Contacts are scored against the canonical name, alias, and phone directly, so an aliased contact still wins on its real name.
- `SearchContacts` now also `LIKE`s on `business_name` and `first_name` so business contacts and first-name-only entries are reachable.
- A DB hit explainable only by a hidden field (`push_name`, `business_name`, `first_name`) falls back to the substring score — unless the query is itself a substring of the JID, which would be noise (e.g. `us` matching every `@g.us`).
- Groups and chats have no hidden textual fields, so a zero visible rank means the query matched only the JID and the row is dropped.
- The resolver fetches every match the store returns before ranking, so an older but exactly-named chat can't be truncated away behind newer substring hits (`ListChats` orders by recency).

## Files

- `internal/resolve/` — new package, pure-Go ranked resolver with a small `Source` interface and unit tests. No new dependencies.
- `cmd/wacli/recipient.go` — CLI wrapper, picker, policy glue.
- `cmd/wacli/helpers.go` — new `isInteractive()` that checks stdin+stderr.
- `cmd/wacli/send.go`, `cmd/wacli/send_file_cmd.go` — flag wiring, help text, `--pick`.
- `internal/store/chats_contacts_groups.go` — expanded `SearchContacts` `WHERE` clause.
- `README.md` — usage snippet.

838 insertions, 13 deletions. No changes to the existing JID/phone code path.

## Test plan

- [x] `go build -tags sqlite_fts5 ./...`
- [x] `go vet -tags sqlite_fts5 ./...`
- [x] `go test -tags sqlite_fts5 ./...` — all packages green.
- [x] New unit tests in `internal/resolve/resolve_test.go` covering: phone/JID detection, phone normalization, ranking precedence, alias vs real-name scoring, numeric group names, DB hits on hidden fields, JID-only noise filtering (`us` → `@g.us`), `Trip 2024` regression, word-boundary matching, empty-input validation.
- [x] Manual end-to-end against a seeded store: JID passthrough, `--to mom` alias exact, `--to john` ambiguous picker, `--to zzz` clean no-match, `+49 170 1234567` phone normalization, `--to 2024` group exact.
- [ ] Interactive picker on a real TTY (the seeded manual run verified the non-TTY / `--json` path; a reviewer on a terminal can confirm the prompt UX).

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,12 @@ pnpm wacli history backfill --chat 1234567890@s.whatsapp.net --requests 10 --cou
 # Download media for a message (after syncing)
 ./wacli media download --chat 1234567890@s.whatsapp.net --id <message-id>
 
-# Send a message
+# Send a message — by JID, phone, or a name from your contacts/groups/chats
 pnpm wacli send text --to 1234567890 --message "hello"
+pnpm wacli send text --to "mom" --message "landed"
+pnpm wacli send text --to "Family" --message "on my way"
+# Ambiguous name? Pick non-interactively with --pick N:
+pnpm wacli send text --to john --pick 2 --message "hi"
 
 # Send a file
 ./wacli send file --to 1234567890 --file ./pic.jpg --caption "hi"
--- a/cmd/wacli/helpers.go
+++ b/cmd/wacli/helpers.go
@@ -13,6 +13,10 @@ func isTTY() bool {
 	return term.IsTerminal(int(os.Stdout.Fd()))
 }
 
+func isInteractive() bool {
+	return term.IsTerminal(int(os.Stdin.Fd())) && term.IsTerminal(int(os.Stderr.Fd()))
+}
+
 func parseTime(s string) (time.Time, error) {
 	s = strings.TrimSpace(s)
 	if s == "" {
--- a/cmd/wacli/recipient.go
+++ b/cmd/wacli/recipient.go
@@ -1,0 +1,142 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/steipete/wacli/internal/app"
+	"github.com/steipete/wacli/internal/resolve"
+	"github.com/steipete/wacli/internal/wa"
+	"go.mau.fi/whatsmeow/types"
+)
+
+type recipientOptions struct {
+	pick   int
+	asJSON bool
+}
+
+func resolveRecipient(a *app.App, input string, opts recipientOptions) (types.JID, error) {
+	input = strings.TrimSpace(input)
+	if input == "" {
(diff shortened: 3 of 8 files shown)


steipete:
Thanks for the PR and for the clear implementation idea. I rebuilt this on current `main` in 56f9c26 because the branch had gone stale against the newer send pipeline and store layout.

The shipped version keeps the current retry/reconnect path, quoted replies, `send react`, file-send behavior, and read-only checks intact, while adding the same user-facing feature: `send text/file --to` can resolve synced contacts, groups, and chats by name, with `--pick N` for non-interactive ambiguity. I also kept the contact first/business-name search improvement, added regression coverage, updated README/spec, and credited you in the changelog plus commit trailer.

Closing this as superseded by the maintainer rebuild now on `main`.

PR #123, opened 2026-04-11
fix message search filters

Mansehej:
## Summary
- add `--has-media` support to `wacli messages search`
- treat `--type text` as messages without media while preserving existing media-type filtering
- add store tests covering the new search filter behavior

### DIFF
--- a/cmd/wacli/messages.go
+++ b/cmd/wacli/messages.go
@@ -122,6 +122,7 @@ func newMessagesSearchCmd(flags *rootFlags) *cobra.Command {
 	var limit int
 	var afterStr string
 	var beforeStr string
+	var hasMedia bool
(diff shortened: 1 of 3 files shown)


ImLukeF:
Review note: `--has-media` combined with `--type text` currently turns into a contradictory filter and silently returns an empty result set. That reads like a bug from the CLI side; the user asked for an impossible combination and should get a clear validation error instead. I verified a local fix that rejects that combination and added a regression test for it. I couldn’t push the patch because GitHub rejected branch updates from my account with `403 Permission denied`.

steipete:
Thanks @Mansehej. The search-filter work is now covered by #128 / 54d44b3. Closing this duplicate as superseded.

PR #125, opened 2026-04-11
feat: filter out left groups from `groups list`

### LINKED ISSUE #23: groups list: filter out groups user has left

SeifBenayed:
## Summary

Closes #23.

Groups the user has left are now excluded from `groups list` by default. A new `--include-left` flag shows all groups including left ones.

## Changes

- **Migration 4**: Adds `left_at INTEGER` column to `groups` table (idempotent)
- **`UpsertGroup()`**: Resets `left_at=NULL` on upsert (group is active if we see it)
- **`MarkGroupLeft(jid)`**: Sets `left_at` to current Unix timestamp
- **`ListGroups(query, limit, includeLeft)`**: Filters `WHERE left_at IS NULL` when `includeLeft=false`
- **`refreshGroups()`**: After fetching joined groups from WhatsApp, marks any DB-only groups as left
- **`groups leave`**: Marks the group as left in DB after successful leave
- **`groups list --include-left`**: New flag (default: false)

## Usage

```bash
# List only active groups (default)
wacli groups list

# Include groups you've left
wacli groups list --include-left
```

## Test plan

- [x] `go build -tags sqlite_fts5 ./cmd/wacli` compiles
- [x] `go test -tags sqlite_fts5 ./...` — all tests pass
- [x] go.mod/go.sum unchanged
- [x] Migration is idempotent (checks column existence before ALTER)

### DIFF
--- a/cmd/wacli/groups_info_rename.go
+++ b/cmd/wacli/groups_info_rename.go
@@ -146,6 +146,7 @@ func newGroupsLeaveCmd(flags *rootFlags) *cobra.Command {
 			if err := a.WA().LeaveGroup(ctx, gjid); err != nil {
 				return err
 			}
+			_ = a.DB().MarkGroupLeft(gjid.String())
 			if flags.asJSON {
 				return out.WriteJSON(os.Stdout, map[string]any{"jid": gjid.String(), "left": true})
 			}
--- a/cmd/wacli/groups_refresh_list.go
+++ b/cmd/wacli/groups_refresh_list.go
@@ -57,6 +57,7 @@ func newGroupsRefreshCmd(flags *rootFlags) *cobra.Command {
 func newGroupsListCmd(flags *rootFlags) *cobra.Command {
 	var query string
 	var limit int
+	var includeLeft bool
 	cmd := &cobra.Command{
 		Use:   "list",
 		Short: "List known groups (from local DB; run sync to populate)",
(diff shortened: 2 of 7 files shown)


ImLukeF:
Review note: the new "mark missing groups as left" pass only inspects the first 50 stored groups. `refreshGroups()` calls `ListGroups("", 0, true)`, but `ListGroups` still treats `limit <= 0` as the default 50-row window, so older groups past that cutoff never get marked left. I verified a local fix by making the zero-limit path fetch all rows and added a regression test with 55 groups to cover it. I couldn’t push the patch because GitHub rejected branch updates from my account with `403 Permission denied`.

steipete:
Thanks @SeifBenayed. Left-group filtering landed on main in 21063d8 and was extended in 95f2136 so refresh marks groups missing from the live joined list as left. Closing as superseded.

PR #127, opened 2026-04-11
[fix] bounded-backpressure media queue for sync

ImLukeF:
This supersedes and fixes the remaining gaps in #121.

The change tightens the media sync path with bounded backpressure so downloads do not pile up unboundedly during sync. That keeps the queue under control while still allowing media to flow through as capacity becomes available.

Summary:
- bound the media queue instead of letting sync enqueue without limit
- preserve sync progress while preventing runaway in-flight media work
- cover the behavior in tests so the queue pressure stays observable

Root cause: the previous flow could accumulate too much media work during sync, which made the queue grow beyond what the consumer path could drain promptly.

### DIFF
--- a/internal/app/fake_wa_test.go
+++ b/internal/app/fake_wa_test.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/steipete/wacli/internal/wa"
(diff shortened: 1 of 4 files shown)


steipete:
Closing as superseded by #121 / 12e8e1a, which added bounded media enqueue backpressure. Thanks @ImLukeF.

PR #128, opened 2026-04-11
fix: reject contradictory message search filters

ImLukeF:
This supersedes and fixes the gaps in #123.

It adds a guard for the invalid `--has-media --type text` combination, and makes message type filtering treat text and media consistently so search results stay correct across the supported filter set.

The change is covered by store tests for `has-media`, `type=text`, `type=image`, and the contradictory combination.


steipete:

- Rewrite commit: c3f6b067439111717a6f8fb6e9361ef5779567c8
- Merge commit: 54d44b34fca51a112fc32e8b7e894f4950f34ed8

Thanks @ImLukeF and @Mansehej!

PR #129, opened 2026-04-11
fix: mark left groups beyond the default list window

ImLukeF:
This supersedes and fixes the gaps in #125.\n\nThe original change handled the left-group state for the groups returned by the default list window, but it did not cover groups beyond that window. This branch removes the hard 50-item cap in the group listing path used by refresh, so older groups can still be discovered and marked left instead of disappearing from the active set.\n\nValidation:\n- refreshGroups now has coverage for a missing-group case beyond the first 50 groups\n- ListGroups no longer forces a default limit when callers pass 0, so refresh can examine the full set when needed


steipete:
Thanks @ImLukeF. The beyond-default-list case landed on main in 95f2136: groups refresh now marks any active local group missing from live joined groups as left. Closing as superseded.

PR #130, opened 2026-04-11
Fix recipient normalization gaps from #116

ImLukeF:
This PR supersedes and fixes the remaining gaps in #116.

It hardens recipient parsing so `+` is only accepted as a leading E.164 marker, and rejects embedded plus signs and non-ASCII digits while still accepting formatted phone numbers and full JIDs.

Tests added in `internal/wa/client_test.go` cover the new rejection cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ pnpm wacli history backfill --chat 1234567890@s.whatsapp.net --requests 10 --cou
 
 # Send a message
 pnpm wacli send text --to 1234567890 --message "hello"
+# phone numbers can also be passed as +E164 or formatted input like "+1 (234) 567-8900"
 
 # Send a file
 ./wacli send file --to 1234567890 --file ./pic.jpg --caption "hi"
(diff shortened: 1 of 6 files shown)


steipete:
Current main already has the hardening half here: plus is only allowed as the first character, Unicode digits are rejected, and the parser enforces 7-15 ASCII digits. What main does not do is accept formatted phone numbers with spaces/dashes/parens. I am leaving this open because that is a product/UX choice, not a proven bug fix.

steipete:
Landed the remaining recipient-normalization part on main in `1e8342f`.

What changed:
- phone-style recipients now accept common formatting: spaces, dashes, parentheses, dots, and a single leading `+`
- the normalized WhatsApp JID user remains digits-only with the existing 7-15 ASCII digit guard
- embedded `+`, letters, Unicode digits, too-short, and too-long values remain rejected
- updated relevant flag help plus README/changelog

Dependency checked: whatsmeow `types.JID` for `s.whatsapp.net` users expects the phone-number user component; wacli still constructs `types.JID{User: <digits>, Server: types.DefaultUserServer}` and only relaxes CLI input normalization before that boundary.

Proof:
- `go test ./cmd/wacli ./internal/wa`
- Full gate: `pnpm format:check && pnpm lint && pnpm test && pnpm build && git diff --check`

Contributor credit preserved in the commit trailer. Closing this PR as superseded by `main`.

PR #126, opened 2026-04-11
fix: restrict file permissions on databases and sensitive files

SeifBenayed:
## Summary

Closes #50.

Ensures all database files, session data, and downloaded media are created with restricted permissions — not world-readable.

## Changes

- **`internal/store/db.go`** — `os.Chmod(path, 0o600)` after DB init (wacli.db)
- **`internal/wa/client.go`** — `os.Chmod(storePath, 0o600)` after session.db creation
- **`internal/wa/media.go`** — `os.Chmod(targetPath, 0o600)` on downloaded media files; standardized dir perms to `0o700`
- **`internal/lock/lock.go`** — Standardized octal notation (`0700` → `0o700`, `0600` → `0o600`)
- **`internal/app/media.go`** — Standardized octal notation for media dir

## Permission model

| Resource | Permission | Rationale |
|----------|-----------|-----------|
| Store directory (`~/.wacli/`) | `0o700` | Only owner can list contents |
| Database files (`.db`) | `0o600` | Contains messages, session keys |
| Lock file | `0o600` | Contains PID |
| Downloaded media | `0o600` | User's private media |
| Media directories | `0o700` | Only owner can traverse |

## Tests added

- `TestDBFilePermissions` — verifies DB files are `0o600` and subdirs are `0o700`
- `TestStoreDirPermissions` — verifies store dir is `0o700` and wacli.db is `0o600`

## Test plan

- [x] `go build -tags sqlite_fts5 ./cmd/wacli` compiles
- [x] `go test -tags sqlite_fts5 ./...` — all tests pass
- [x] go.mod/go.sum unchanged
- [x] No functional behavior change — only permission hardening

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #131, opened 2026-04-11
fix: harden sqlite sidecar permissions

ImLukeF:
This is the repaired follow-up to #126 and supersedes the gaps in that branch.

It hardens SQLite artifacts by forcing owner-only permissions on the main DB file and its `-wal`, `-shm`, and `-journal` sidecars after initialization, regardless of umask.

Tests now exercise a write path and verify the permissions on any sidecar files that appear.

### DIFF
--- a/internal/app/app_test.go
+++ b/internal/app/app_test.go
@@ -1,8 +1,39 @@
 package app
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 )
+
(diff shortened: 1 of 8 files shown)


steipete:
Closing as superseded: database file and directory permission hardening is already on main via #147 plus follow-up chmod coverage. Thanks @ImLukeF.

PR #144, opened 2026-04-11
security: validate phone numbers in ParseUserOrJID (#55)

### LINKED ISSUE #55: security: minimal input validation on phone numbers / JIDs
## Description

In `ParseUserOrJID` (`internal/wa/client.go:250-259`), when input doesn't contain `@`, it's treated as a raw phone number and used directly as the JID User field with no validation of format, length, or character set.

**Severity:** Medium

**File:** `internal/wa/client.go:250-259`

## Current Behavior

```go
return types.JID{User: s, Server: types.DefaultUserServer}, nil
```

Any arbitrary string is accepted.

## Expected Behavior

Validate that phone numbers match E.164 format (1-15 digits).

## Suggested Fix

Add regex validation: `^\d{1,15}$`. Fix submitted in PR from `alexander-morris/wacli:security/audit-fixes`.

draix:
## Summary

`ParseUserOrJID` accepted any string as a raw phone number and used it directly as the JID `User` field with no validation. Invalid inputs like `123abc` or a 2-digit string were silently forwarded to WhatsApp servers, where they caused confusing `info query timed out` errors (visible in issues #28, #44, #91).

## Changes

When the input does not contain `@` (i.e. it is treated as a phone number), validate:
- **Length**: 7–15 digits (ITU-T E.164 range)
- **Charset**: digits only, after stripping an optional leading `+`

Invalid inputs are rejected immediately with a clear error message before any network call is made.

## Testing

- 9 new sub-tests covering valid numbers (+prefix, min length, max length) and invalid ones (too short, too long, non-digit chars, leading + with letters, empty string).
- All existing `ParseUserOrJID` tests pass.
- Verified against a live WhatsApp account: `wacli send text --to 123abc ...` now returns `invalid phone number "123abc": unexpected character 'a'` instead of timing out after 5s.

Closes #55

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 ### Security
 
 - Auth: reject `?` and `#` in whatsmeow session store paths to avoid SQLite URI parameter injection. (#180 — thanks @shaun0927)
+- Send: validate phone-number recipients before constructing WhatsApp JIDs. (#144 — thanks @draix)
 - Store: restrict index and session SQLite database files to owner-only permissions. (#147 — thanks @draix)
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


steipete:

- Rewrite commit: baad358719942098a35c247712d31cb2a49c7f6a
- Merge commit: 318421d4158f1e01ccc8340f08a09bb6dcded48e

Thanks @draix!

PR #63, opened 2026-02-19
security: harden input validation, resource limits, and file permissions

alexander-morris:
## Summary

Comprehensive security audit with fixes for 12 findings across input validation, resource consumption, file permissions, and concurrency safety.

### Changes

**Resource Limits (HIGH):**
- Add 100MB file size check before reading uploads into memory (`send_file.go`)
- Add 100MB media download size limit to prevent disk exhaustion (`wa/media.go`)

**File Permissions (MEDIUM):**
- Enforce `0600` permissions on SQLite database files after creation (`store/db.go`)
- Reject store paths containing URI-special characters `?` and `#` (`store/db.go`, `wa/client.go`)

**Input Validation (MEDIUM):**
- Validate phone numbers against E.164 format (`^\d{1,15}$`) in `ParseUserOrJID` (`wa/client.go`)

**SQL Safety (LOW):**
- Escape LIKE metacharacters (`%`, `_`) in all search queries with `ESCAPE '\'` (`store/search.go`, `store/chats_contacts_groups.go`)
- Sanitize FTS5 queries by wrapping in double-quotes to prevent operator injection (`store/search.go`)
- Validate table names against `^[a-zA-Z_][a-zA-Z0-9_]*$` in PRAGMA queries (`store/migrations.go`)

**Path Safety (LOW):**
- Strip null bytes and control characters in `SanitizeSegment` and `SanitizeFilename` (`pathutil/sanitize.go`)

**Concurrency & Stability (MEDIUM/LOW):**
- Add `defer recover()` in event handler and media worker goroutines (`app/sync.go`, `app/media.go`)
- Fix goroutine leak: drop media jobs when queue full instead of spawning unbounded goroutines (`app/sync.go`)
- Use `sync.Once` to prevent race condition in `OpenWA` initialization (`app/app.go`)
- Add forced exit on second SIGINT during sync (`cmd/wacli/sync.go`)

### Related Issues

Closes #50, closes #51, closes #52, closes #55, closes #56, closes #57, closes #58, closes #59, closes #60, closes #61, closes #62

Issues #53 (rate limiting) and #54 (unbounded history sync) are documented but require more extensive changes beyond this PR.

## Test plan

- [ ] Verify `wacli send file` rejects files > 100MB
- [ ] Verify media downloads are rejected when `fileLength` > 100MB
- [ ] Verify database files are created with `0600` permissions
- [ ] Verify phone number validation rejects non-digit input
- [ ] Verify LIKE search with `%` or `_` returns literal matches
[...]

PR #147, opened 2026-04-11
security: restrict database file permissions to owner-only (0600) (#50)

### LINKED ISSUE #50: security: no explicit file permissions on session and application databases
## Description

While the parent directories are created with `0700` permissions, the `session.db` and `wacli.db` files themselves rely on SQLite driver defaults (typically `0644` before umask). On multi-user systems with permissive umask, these files could be world-readable.

**Severity:** Medium

**Files:** `internal/wa/client.go:49`, `internal/store/db.go:27`

## Current Behavior

Database files inherit permissions from the SQLite driver default and system umask.

## Expected Behavior

Database files should be explicitly set to `0600` after creation since they contain sensitive data (session keys, message content, phone numbers).

## Suggested Fix

Call `os.Chmod(path, 0600)` after database open. A fix has been submitted in PR from `alexander-morris/wacli:security/audit-fixes`.

draix:
## Summary

SQLite creates database files with permissions determined by the process `umask`. With the common default `umask 0022`, both `wacli.db` and `session.db` are created as **world-readable** (`-rw-r--r-- 0644`). On multi-user systems this exposes WhatsApp session keys and message history to other local users.

The WAL (`-wal`) and SHM (`-shm`) sidecar files — which contain in-progress write data — were also left with `umask`-determined permissions.

## Changes

- In `store/db.go`: call `chmodDB(path, 0o600)` after `init()` (which runs the PRAGMAs that trigger file creation). A new `chmodDB` helper applies `0o600` to the main file and, if they exist, to the `-wal` and `-shm` sidecars.
- In `wa/client.go`: apply the same `0o600` chmod to `session.db` and its sidecars both before and after `sqlstore.New` opens it, covering both existing files (pre-open) and freshly-created ones (post-open).

## Testing

- New `TestDBFilePermissions` integration test: opens a store under `umask 0022` and asserts the resulting file permissions are `0600`, not the `0644` that SQLite would create without the explicit chmod.
- Verified on macOS: `ls -la ~/.wacli/*.db` now shows `-rw-------` for all database files.

## Related

Supersedes part of #63 and #126 (which also address file permissions but with a slightly different scope).

Closes #50

### DIFF
--- a/internal/store/db.go
+++ b/internal/store/db.go
@@ -29,12 +29,37 @@ func Open(path string) (*DB, error) {
 		return nil, fmt.Errorf("open sqlite: %w", err)
 	}
 
+	// Run init (which executes PRAGMAs) before chmod so the driver has
+	// already created the file and any WAL/SHM sidecars.
 	s := &DB{path: path, sql: db}
 	if err := s.init(); err != nil {
 		_ = db.Close()
 		return nil, err
 	}
+
+	// Restrict the DB file and its WAL/SHM sidecars to owner-only (0600)
+	// regardless of umask. On multi-user systems the default sqlite3 driver
+	// permissions (governed by umask, typically 0644) would make session data
+	// world-readable. The sidecars contain in-progress write data and must be
+	// treated with the same sensitivity as the main file (#50).
+	if err := chmodDB(path, 0o600); err != nil {
+		_ = db.Close()
+		return nil, err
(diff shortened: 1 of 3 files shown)


steipete:

steipete:
Closing as already landed on main in 716e7a8 (#147). Thanks @draix.

PR #149, opened 2026-04-11
feat: enrich doctor and auth status with store stats and linked JID

draix:
## Summary

`wacli doctor` and `wacli auth status` provide useful diagnostics but omit information that operators and scripts rely on daily.

### `wacli doctor`

**Before:**
```
STORE          /Users/alice/.wacli
LOCKED         false
AUTHENTICATED  true
CONNECTED      false
FTS5           false
```

**After:**
```
STORE          /Users/alice/.wacli
LOCKED         false
AUTHENTICATED  true
LINKED_JID     15551234567@s.whatsapp.net
CONNECTED      false
FTS5           false
MESSAGES       1 284
CHATS          47
CONTACTS       203
LAST_SYNC      2026-04-11T16:46:39Z
```

Corrupted or missing stores no longer cause a hard exit — doctor shows STORE/LOCKED output and prints a clear error:
```
ERROR: store could not be opened: file is not a database
Tip: check that the store directory exists and is not corrupted.
```

### `wacli auth status`

**Before:** `Authenticated.`

**After:** `Authenticated as 15551234567@s.whatsapp.net`

**`--json` before:** `{"authenticated": true}`

**`--json` after:**
```json
{
  "authenticated": true,
  "linked_jid": "15551234567@s.whatsapp.net",
  "phone": "15551234567"
}
```

## Implementation

- `DB.Stats()` — single SQL query for message/chat/contact counts and latest message timestamp
- `WAClient.LinkedJID()` — returns the non-AD JID string from the device store
- Doctor now opens app in best-effort mode (degraded output rather than hard exit on store errors)

## Testing

- Verified against a real WhatsApp account (90 messages, 50 chats, 50 contacts)
- Corrupted store: shows error without crashing ✅
- Empty store: shows 0 counts without error ✅  
- Locked store: correctly shows LOCKED=true while sync is running ✅
- `--json` output is backward-compatible (new fields are additive)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - CLI: add `--full` to disable table truncation; piped output now keeps full message IDs. (#13 — thanks @rickhallett)
 - CLI: add `presence typing` and `presence paused` commands for WhatsApp composing indicators. (#76 — thanks @redemerco)
+- Diagnostics: show linked JID and local store counts in `auth status` and `doctor`. (#149 — thanks @draix)
 - Messages: add `messages list --sender`, `--from-me`, `--from-them`, and `--asc` filters. (#153 — thanks @draix)
 - Messages: add `messages search --has-media`, `--type text`, case-insensitive media types, and validation for contradictory filters. (#128 — thanks @ImLukeF and @Mansehej)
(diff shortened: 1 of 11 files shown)


steipete:

- Rewrite commit: 6a5c32f4577329b22a9ba2806419fabefe20d039
- Merge commit: ec608225f93f2fafb93c91080da2d4c056f67951

Thanks @draix!

PR #154, opened 2026-04-11
feat: add --reply-to flag to 'wacli send text' for quoted replies

draix:
## Summary

Quoted/threaded replies are the most basic building block for conversational bots and AI agent integrations. Without `--reply-to`, every wacli message arrives as a standalone message with no conversation threading — making it hard for users to track which bot response belongs to which question.

## Usage

```bash
# Quote a message in a DM
wacli send text --to +15551234567 --message "Got it!" --reply-to 3EB0A3DCA3175E17850852

# Quote a message in a group (--reply-to-sender required)
wacli send text --to 1234567890-123456789@g.us \
  --message "Agreed with @Alice" \
  --reply-to 3EB0A3DCA3175E17850852 \
  --reply-to-sender +15559876543
```

## Flags added

```
--reply-to          message ID to quote/reply to
--reply-to-sender   JID of the quoted message's author (required for group replies)
```

## Implementation

Add `SendTextReply` to `wa.Client` using whatsmeow's `ExtendedTextMessage` + `ContextInfo{StanzaID, Participant}`. When `--reply-to` is present, routes through `SendTextReply`; otherwise falls through to the existing `SendText`.

## Testing

Tested against a real WhatsApp account — quoted reply delivered correctly with the green quote bubble showing the original message.

Used in production at [Zapia](https://zapia.ai) for AI agent responses that quote the user's question.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Messages: show display text for replies, reactions, and media in `messages context`. (#183 — thanks @fuleinist)
 - Send: strip a leading `+` from phone-number recipients before building WhatsApp JIDs. (#74 — thanks @FrederickStempfle)
 - Search: keep FTS5 enabled after reopening existing databases with already-applied migrations. (#185 — thanks @iamhitarth)
+- Send: add `send text --reply-to` for quoted replies, with sender inference for synced group messages. (#154 — thanks @draix)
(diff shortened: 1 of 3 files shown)


steipete:

- Land commit: 782c290

Thanks @draix!

Issue #36, opened 2026-02-10
Suggestion: Add WACLI_STORE_DIR environment variable to configure store dir


jsmouret:
You are awesome :)
Thank you!

PR #155, opened 2026-04-11
[codex] Add store-dir env override and Windows locking

TerminalsandCoffee:
## Summary
- add `WACLI_STORE_DIR` as a default store-directory override, while keeping `--store` highest priority
- document the new store-dir override in the CLI help, README, and spec
- split file locking into Unix and Windows implementations so the CLI compiles and enforces single-instance locking on Windows

## Why
- issue #36 asks for an environment-based store-dir override
- while validating the change on Windows, the current lock implementation failed to compile because it depends on Unix-only `syscall.Flock`

## Validation
- `go test ./internal/config ./internal/lock`
- `go test ./cmd/wacli -run TestDoesNotExist`

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Run (local build only):
 
 ## Quick start
 
-Default store directory is `~/.wacli` (override with `--store DIR`).
+Default store directory is `~/.wacli` (override with `--store DIR` or `WACLI_STORE_DIR`).
 
 ```bash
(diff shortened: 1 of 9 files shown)


steipete:
Thanks @TerminalsandCoffee. The store-dir env override and Windows lock split are now on main via the 0.6/0.7 work, including WACLI_STORE_DIR and Windows lock compile coverage. Closing as superseded.

PR #165, opened 2026-04-15
fix: escape LIKE wildcards in ListChats, ListGroups, and SearchContacts

hobostay:
## Summary

- **Fix LIKE wildcard injection in 3 search functions**: `ListChats`, `ListGroups`, and `SearchContacts` built their SQL `LIKE` needles directly from user input (`"%" + query + "%"`) without escaping `%` and `_` wildcards. The `escapeLIKE()` helper already existed and was correctly used in `searchLIKE()`, but these three functions were missed. A user searching for `"100%"` would match any chat with `"100"` followed by any characters, rather than only chats literally containing `"100%"`.
- **Fix `truncate()` to handle multi-byte UTF-8 correctly**: The function used byte-based slicing (`len(s)`, `s[:max]`) which could split multi-byte characters (Chinese, emoji, etc.) and produce invalid UTF-8 output. Changed to rune-based slicing.

### LIKE wildcard escape
- `escapeLIKE()` is now applied in `ListChats`, `ListGroups`, and `SearchContacts` — same as `searchLIKE()`
- `ESCAPE '\'` clause added to all LIKE predicates for consistency
- This is consistent with the existing fix in #145 (`ff2d74d`) which fixed `searchLIKE`

### truncate UTF-8 fix
- `len(s)` → `len(runes)` where `runes = []rune(s)`
- `s[:max]` → `string(runes[:maxRunes])`
- Parameter renamed from `max` to `maxRunes` for clarity

## Test plan

- [x] Added `TestListChatsWildcardEscape` — verifies `%` and `_` in chat search queries are treated as literals
- [x] Added `TestListGroupsWildcardEscape` — verifies `%` and `_` in group search queries are treated as literals  
- [x] Added `TestSearchContactsWildcardEscape` — verifies `%` and `_` in contact search queries are treated as literals
- [x] All existing tests pass (`go test ./...`)
- [x] Build succeeds (`go build ./...`)

### DIFF
--- a/cmd/wacli/helpers.go
+++ b/cmd/wacli/helpers.go
@@ -27,14 +27,15 @@ func parseTime(s string) (time.Time, error) {
 	return time.Time{}, fmt.Errorf("unsupported time format %q (use RFC3339 or YYYY-MM-DD)", s)
 }
 
-func truncate(s string, max int) string {
+func truncate(s string, maxRunes int) string {
 	s = strings.ReplaceAll(s, "\n", " ")
 	s = strings.TrimSpace(s)
-	if max <= 0 || len(s) <= max {
+	runes := []rune(s)
+	if maxRunes <= 0 || len(runes) <= maxRunes {
 		return s
 	}
-	if max <= 1 {
-		return s[:max]
+	if maxRunes <= 1 {
+		return string(runes[:maxRunes])
 	}
-	return s[:max-1] + "…"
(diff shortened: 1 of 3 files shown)


steipete:
Closing as fixed on main: LIKE patterns now escape wildcards before ListChats/ListGroups/SearchContacts matching. Thanks @hobostay.

Issue #169, opened 2026-04-17
fix: sync --once hangs indefinitely when no events are received after connect


fuleinist:
file:///home/admin/.openclaw/workspace/tmp-pr-body.txt

steipete:
Fixed on main: sync --once now exits after idle and message-event activity resets that timer, so it no longer hangs forever on quiet sessions. Thanks for reporting.

PR #171, opened 2026-04-17
fix: reset idle timer on Connected event in sync --once mode

fuleinist:
## Problem

When running `wacli sync --once`, the process hangs indefinitely on subsequent syncs when no new messages arrive after connection.

The root cause: the idle timer (`--idle-exit`) was only reset when events like `Message` or `HistorySync` were received. On already-synced accounts, no such events arrive after `Connected`, so the timer incorrectly uses the time when `Sync()` was called instead of when the connection was established.

This is tracked in:
- Issue #87: sync --once hangs indefinitely at "Connected." when no history sync event is received
- Issue #169: sync --once hangs indefinitely when no events are received after connect

## Solution

Explicitly reset `lastEvent` when the `Connected` event is received:

```go
case *events.Connected:
    fmt.Fprintln(os.Stderr, "\nConnected.")
    lastEvent.Store(time.Now().UTC().UnixNano())
```

This ensures the idle timer correctly counts from connection time, not from when the sync process started.

## Testing

- First sync: `Connected` then `HistorySync` arrives - timer reset correctly by HistorySync (no regression)
- Subsequent syncs: `Connected` but no new events - timer now correctly starts from connection time

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Messages: show display text for replies, reactions, and media in `messages context`. (#183 — thanks @fuleinist)
 - Search: keep FTS5 enabled after reopening existing databases with already-applied migrations. (#185 — thanks @iamhitarth)
 - Send: persist retry-message plaintext so linked devices can decrypt retried messages. (#186 — thanks @SimDamDev)
+- Sync: start `sync --once` idle timing after the `Connected` event. (#171 — thanks @fuleinist)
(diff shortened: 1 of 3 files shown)


steipete:
Landed via rebase in ee3f28bb12af1260253dd885ce8ac29c5e543b4c. Reworked this into a regression around delayed connect/idle timing and kept the handler implementation single-source. Full gate passed locally, and GitHub checks passed before merge. Thanks @fuleinist.

PR #172, opened 2026-04-17
fix(config): use XDG state dir on Linux by default

### LINKED ISSUE #164: XDG instead of cluttering home directory
Please consider using the XDG Base Directory specification https://specifications.freedesktop.org/basedir/latest/ instead of just storing stuff directly under $HOME.

So that instead of ~/.wacli, you'd use ~/.config/wacli , ~/.local/share and ~/.local/state . And of course overriding would be possible via environment.

txhno:
Closes #164

## Summary
- use `${XDG_STATE_HOME}/wacli` or `~/.local/state/wacli` as the default store on Linux
- keep `WACLI_STORE_DIR` and `--store` as explicit overrides
- update help text, tests, and docs to match the new default path

## Testing
- `go test ./internal/config ./cmd/wacli`
- `go test ./...`


steipete:
Thanks @txhno. Reimplemented this on main as fa1c800 with the XDG Linux default plus a legacy ~/.wacli fallback when an existing store is present, and docs/tests updated in 8a0e13b. Closing as superseded.

PR #179, opened 2026-04-17
bug: recover per media job so a panic no longer kills the worker (#176)

### LINKED ISSUE #176: bug: media worker panic recovery placed outside for-loop drains pool (regression of #52/#143)
## Description

PR #143 added panic recovery to the media worker goroutines at `internal/app/media.go:86-110`, but the `defer recover()` is registered **outside** the `for { select { ... } }` loop. When `downloadMediaJob` panics, the deferred recover fires, prints a line to stderr, and the goroutine returns. Each panicking job permanently shrinks the worker pool by one; if all workers are lost, media downloads stall silently even though the rest of the sync continues and the application never exits.

**Severity:** Medium

**File:** `internal/app/media.go:86-110`

## Current Behavior

A panic inside `downloadMediaJob` unwinds past the `for` loop, triggers the deferred recover at the goroutine top, and the goroutine returns. The worker is not respawned. Four workers after four panicking jobs → zero workers left.

## Expected Behavior

Panic recovery should wrap each per-job invocation so a bad payload fails only that job and the worker keeps consuming the queue.

## Suggested Fix

Move the `defer recover()` inside the loop body, wrapping the `downloadMediaJob` call in an inner function. Capture the offending `chatJID`/`msgID` (and a stack trace) in the log line so panics remain diagnosable after the fix.

shaun0927:
## Summary

PR #143 added `recover()` to the media worker goroutines, but the `defer` is registered at the goroutine top rather than per-job. A panic inside `downloadMediaJob` unwinds past the `for { select }` loop, fires the deferred recover, and the goroutine returns — so one bad payload permanently removes a worker from the pool. Under sustained malformed input the pool drains to zero and media downloads stall silently while the rest of the process keeps running.

## Changes

- Move the recover `defer` into an inner function that wraps each `downloadMediaJob` call so the `for { select }` loop continues to consume the queue after a panic.
- Include `chatJID`, `msgID`, and `runtime/debug.Stack()` in the recovered-panic log line so panics remain diagnosable.

## Testing

- New test `TestMediaWorkerSurvivesPanic` in `internal/app/media_panic_test.go` wraps `newFakeWA()` with a stub whose `DownloadMediaToFile` panics on the first invocation. It spawns a 1-worker pool, feeds it a panicking job followed by a benign one, and asserts both calls reached the download path and the benign job persisted a `LocalPath` — proving the worker survived the panic.
- Existing `TestDownloadMediaJobMarksDownloaded` still passes.
- `go test -tags sqlite_fts5 -race ./internal/app/ ./internal/wa/ ./internal/store/` is green.

Closes #176.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 
+- Media: recover panics per download job so one bad payload no longer drains the worker pool. (#179 — thanks @shaun0927)
 - Send: persist retry-message plaintext so linked devices can decrypt retried messages. (#186 — thanks @SimDamDev)
 
 ### Docs
--- a/internal/app/media.go
+++ b/internal/app/media.go
@@ -92,7 +92,10 @@ func (a *App) runMediaWorkers(ctx context.Context, jobs <-chan mediaJob, workers
 				select {
 				case <-ctx.Done():
 					return
-				case job := <-jobs:
+				case job, ok := <-jobs:
+					if !ok {
+						return
+					}
 					if strings.TrimSpace(job.chatJID) == "" || strings.TrimSpace(job.msgID) == "" {
 						continue
 					}
--- a/internal/app/media.go
+++ b/internal/app/media.go
@@ -7,6 +7,7 @@ import (
 	"mime"
 	"os"
 	"path/filepath"
+	"runtime/debug"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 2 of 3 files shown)


steipete:
Landed via rebase onto main.

Kept the per-job panic regression test and added a closed-channel exit guard in the worker loop.

Thanks @shaun0927!

PR #181, opened 2026-04-17
enhance: log stack trace, event type, and counter when recovering sync panics (#178)

### LINKED ISSUE #178: enhancement: capture stack trace and counter on recovered panics (follow-up to #52/#143)
## Description

The panic recovery added in #143 logs only `fmt.Fprintf(os.Stderr, "... (recovered): %v\n", r)` in both `internal/app/sync.go:86-89` and `internal/app/media.go:92-96`. In practice this is indistinguishable from a silent swallow:

- No `runtime/debug.Stack()` — the panic location is unknown.
- No counter — a one-off bad payload looks the same as a pathological loop.
- No context — nothing identifies which event type / message ID / media file triggered the panic.

wacli's stated threat model (best-effort sync over an external protocol) puts malformed server payloads squarely in scope: they are exactly what `recover()` is supposed to catch. Without telemetry on recoveries those bugs never surface for the maintainers to fix.

**Severity:** Low (enhancement on top of #143)

**Files:** `internal/app/sync.go:86-89`, `internal/app/media.go:92-96`

## Proposed Enhancement

1. Log `runtime/debug.Stack()` alongside the recover value.
2. Track a simple `atomic.Int64` panic counter; log the running total on each event.
3. Include available context (event type, `chatJID`, `msgID`, filename) in the log line.
4. Optional: add a soft circuit breaker that surfaces a WARNING (or exits non-zero under `--strict`) if the panic rate exceeds a threshold (e.g. 10 / 60 s).

This is purely additive — no user-visible behavior change on the happy path. Test coverage can reuse the reproducer from any follow-up to #52/#143: stub a handler that panics, assert the stderr output contains the counter and a stack frame.

shaun0927:
## Summary

The event-handler recovery added in #143 logs only the bare recover value at `internal/app/sync.go:86-89`:

```go
fmt.Fprintf(os.Stderr, "\nevent handler panic (recovered): %v\n", r)
```

In practice this is indistinguishable from a silent swallow — no stack trace, no event context, no way to tell a one-off bad payload from a pathological loop that recovers thousands of times per minute. Because wacli's threat model explicitly puts malformed WhatsApp payloads in scope (that is exactly why #143 was added), recoveries need to carry enough information for maintainers to act on.

This PR adds three pieces of telemetry to the existing recovery block:

- `runtime/debug.Stack()` so the panic location is always attached to the log line.
- `%T` of the event so the event class (e.g. `*events.Message`, `*events.HistorySync`) is visible without parsing the stack.
- A per-`Sync` `atomic.Int64` counter appended to the log line (`total=N`) so operators can tell whether recoveries are rare or continuous.

The counter is scoped to each `Sync` invocation on purpose — it doesn't need to be process-global and keeping it in the closure avoids leaking state between sync sessions.

## Changes

- `internal/app/sync.go`: capture `runtime/debug.Stack()`, `%T`, and an `atomic.Int64` counter inside the existing `defer recover()` block. No change to behavior on the happy path.

## Testing

- New test `TestSyncEventHandlerPanicHasStackAndCounter` in `internal/app/sync_panic_test.go` captures `os.Stderr`, emits two typed-nil `*events.Message` values (which force `wa.ParseLiveMessage` to panic on the first dereference), runs a short `Sync`, and asserts the recovery log contains `total=1`, `total=2`, `event=*events.Message`, and a stack frame.
- `go test -tags sqlite_fts5 -race ./internal/app/ ./internal/wa/ ./internal/store/` is green.

This PR is intentionally scoped to `sync.go` because the sibling recover in `internal/app/media.go` is already being reworked in PR #179 (media worker pool regression). Mirroring the same telemetry into `media.go` can follow once #179 lands, to avoid conflicting diffs.

Closes #178.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Media: recover panics per download job so one bad payload no longer drains the worker pool. (#179 — thanks @shaun0927)
 - Send: persist retry-message plaintext so linked devices can decrypt retried messages. (#186 — thanks @SimDamDev)
+- Sync: include event type, stack trace, and recovery count when logging recovered event-handler panics. (#181 — thanks @shaun0927)
 
 ### Docs
 
--- a/internal/app/sync.go
+++ b/internal/app/sync.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"runtime/debug"
 	"strings"
 	"sync/atomic"
 	"time"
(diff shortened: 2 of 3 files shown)


steipete:
Landed via rebase onto main.

Kept the panic log regression test.

Thanks @shaun0927!

Issue #173, opened 2026-04-17
bug: messages context uses raw Text instead of DisplayText, showing blank for reactions/replies


kuishou68:
## Bug Report

**File:** `cmd/wacli/messages.go`
**Command:** `wacli messages context`

### Description

The `messages context` command renders messages using `m.Text` directly:

```go
line := m.Text
if m.MsgID == id {
    line = ">> " + line
}
```

However, `messages list` correctly prefers `m.DisplayText` first and falls back to `m.Text`:

```go
text := strings.TrimSpace(m.DisplayText)
if text == "" {
    text = strings.TrimSpace(m.Text)
}
if m.MediaType != "" && text == "" {
    text = "Sent " + m.MediaType
}
```

### Impact

For messages that only have `DisplayText` set (reactions, reply-quotes, media messages where `Text` is empty), the `context` command shows a blank line, while `list` shows the correct text. This makes `context` output confusing and incomplete.

### Fix

Align `messages context` to use the same display-text priority as `messages list`:

```go
line := strings.TrimSpace(m.DisplayText)
if line == "" {
    line = strings.TrimSpace(m.Text)
}
if m.MediaType != "" && line == "" {
    line = "Sent " + m.MediaType
}
```

Signed-off-by: Cocoon-Break <54054995+kuishou68@users.noreply.github.com>

kuishou68:
I've opened PR #175 to fix this issue.

steipete:
Fixed on main: messages context now uses DisplayText for readable output. Thanks for reporting.

PR #182, opened 2026-04-17
fix: use DisplayText in messages context command

fuleinist:
## Fix for Issue #173

The `messages context` command was using `m.Text` directly, which shows blank for reactions, replies, and media messages where DisplayText should be used instead.

### Changes

Align `messages context` with `messages list` by using the same display-text priority:
1. Prefer `DisplayText` over `Text`
2. Fall back to `Text` if `DisplayText` is empty
3. Show "Sent <mediatype>" for media messages with no text

This matches the existing behavior in `messages list` and fixes the confusing output where context would show blank lines for reactions/replies.

Signed-off-by: fuleinist <fuleinist@outlook.com>

### DIFF
--- a/cmd/wacli/messages.go
+++ b/cmd/wacli/messages.go
@@ -311,7 +311,13 @@ func newMessagesContextCmd(flags *rootFlags) *cobra.Command {
 				if m.FromMe {
 					from = "me"
 				}
-				line := m.Text
+				line := strings.TrimSpace(m.DisplayText)
+				if line == "" {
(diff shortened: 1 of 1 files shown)


steipete:
This is covered by #183, landed in a022c493990da73cc5693c0a3ba5fa5454bf016c with tests for display text, text fallback, and media fallback. Thanks @fuleinist.

steipete:
Closing as fixed on main: messages context now uses DisplayText for human-readable history output. Thanks @fuleinist.

PR #183, opened 2026-04-17
messages context: prefer DisplayText over raw Text (fix #173)

### LINKED ISSUE #173: bug: messages context uses raw Text instead of DisplayText, showing blank for reactions/replies

fuleinist:
fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - Media: recover panics per download job so one bad payload no longer drains the worker pool. (#179 — thanks @shaun0927)
+- Messages: show display text for replies, reactions, and media in `messages context`. (#183 — thanks @fuleinist)
 - Send: persist retry-message plaintext so linked devices can decrypt retried messages. (#186 — thanks @SimDamDev)
 - Sync: include event type, stack trace, and recovery count when logging recovered event-handler panics. (#181 — thanks @shaun0927)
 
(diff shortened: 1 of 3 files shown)


steipete:
Landed via rebase onto main.

I factored the display selection into a helper and added tests for display text, raw-text fallback, and media fallback.

Thanks @fuleinist!

PR #184, opened 2026-04-17
feat(auth): add --phone flag for remote pairing without QR code

bcharleson:
## Problem

`wacli auth` currently only supports QR-code pairing, which requires a terminal that can render a QR image and a phone with a camera pointed at it. This makes `wacli` hard to bootstrap on headless boxes (droplets, CI runners, agent infrastructure) where you only have SSH access.

## Change

Adds a `--phone` flag that uses whatsmeow's `PairPhone` API instead of the QR channel:

```
wacli auth --phone +521234567890
```

The CLI prints an 8-digit code. On the phone, the user opens WhatsApp → Settings → Linked Devices → Link a Device → "Link with phone number instead" and enters the code. The command then polls `IsAuthed()` every 2s for up to 5 minutes and prints success (JSON-aware under ` --json`).

Zero change to the existing QR flow — `wacli auth` with no flag behaves exactly as before.

## Why

I've been running ` wacli` on a remote droplet to back a WhatsApp agent and there was no reasonable way to pair it short of port-forwarding a terminal. This has been in production on my fork for about a week without issues.

## Files

- ` cmd/wacli/auth.go` — new ` --phone` path in the auth command
- ` internal/wa/client.go` — ` Client.PairPhone` method wrapping ` whatsmeow.PairPhone`
- ` internal/app/app.go` — ` PairPhone` added to the ` WAClient` interface
- ` internal/app/fake_wa_test.go` — stub for the new interface method so the app tests keep compiling

## Tests

` go vet ./...`  and ` go test ./...` pass. The new code is a thin wrapper around whatsmeow's own ` PairPhone`, which is already battle-tested upstream.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/cmd/wacli/auth.go
+++ b/cmd/wacli/auth.go
@@ -16,10 +16,11 @@ func newAuthCmd(flags *rootFlags) *cobra.Command {
 	var follow bool
 	var idleExit time.Duration
 	var downloadMedia bool
+	var phone string
 
 	cmd := &cobra.Command{
 		Use:   "auth",
-		Short: "Authenticate with WhatsApp (QR) and bootstrap sync",
+		Short: "Authenticate with WhatsApp (QR or phone number) and bootstrap sync",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx, stop := signalContext()
 			defer stop()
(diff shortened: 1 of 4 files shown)


steipete:
Closing as superseded by focused main patch: `e2bebf6` (`feat: support phone pairing auth`).

What landed:
- `wacli auth --phone <number>` requests WhatsApp phone-number pairing and prints the pairing code to stderr.
- The implementation follows the current whatsmeow contract: connect, wait for the first QR channel code event so the websocket is ready, then call `PairPhone`.
- Existing QR auth stays unchanged when `--phone` is omitted.

Dependency proof: current whatsmeow `PairPhone` source explicitly says the client must be connected normally first and recommends waiting for the first `GetQRChannel` item before calling it. The main patch does that rather than calling `PairPhone` immediately.

Verified:
- `go test ./cmd/wacli ./internal/app ./internal/wa`
- `go test -tags sqlite_fts5 ./cmd/wacli ./internal/app ./internal/wa`

Thanks for the headless-auth PR and production notes.

Issue #160, opened 2026-04-13
v0.5.0 macOS release tarball missing sqlite_fts5 build tag — FTS5 search disabled


jacketyjax:
## Summary

The v0.5.0 universal macOS binary (`wacli-darwin-universal.tar.gz`) was compiled without the `sqlite_fts5` Go build tag, so FTS5 full-text search is silently disabled at runtime:

```
$ wacli doctor
FTS5  false

$ wacli messages search vibecode
Note: FTS5 not enabled; search is using LIKE (slow).
```

## Expected behavior

FTS5 should be enabled in the release artifact, matching the behavior of the Linux build path (which already passes `-tags sqlite_fts5` in the formula).

## Workaround

Build from source via Homebrew HEAD:

```
brew uninstall wacli
brew install --HEAD steipete/tap/wacli
```

After rebuilding, `wacli doctor` reports `FTS5 true`.

## Context

- Confirmed on macOS (Apple Silicon, arm64) with the precompiled tarball from https://github.com/steipete/wacli/releases/tag/v0.5.0
- The `messages_fts*` virtual tables in `wacli.db` still exist but cannot be queried by the FTS5-disabled binary
- Likely just a missing `-tags sqlite_fts5` flag in the release workflow's macOS build step

## References

- Release: https://github.com/steipete/wacli/releases/tag/v0.5.0
- Homebrew tap formula: https://github.com/steipete/homebrew-tap/blob/HEAD/Formula/wacli.rb

PR #185, opened 2026-04-18
fix: detect FTS5 on existing databases after migration already applied

iamhitarth:
## Summary

- Fixes FTS5 detection on databases where migration 3 (`messages fts`) has already been applied
- `doctor` incorrectly reports `FTS5 false` and `messages search` falls back to slow `LIKE` scanning, even when the `messages_fts` table and triggers are fully functional

## Root cause

`migrateMessagesFTS()` sets `d.ftsEnabled = true` only when migration 3 **first runs**. On subsequent DB opens, `ensureSchema()` sees migration 3 is already applied and skips it — so `ftsEnabled` stays at its zero value (`false`).

## Fix

Add a post-migration probe in `init()` that checks if `messages_fts` exists and is queryable, setting `ftsEnabled = true` independently of migration state.

## Related

Partially addresses #160 — the build-tag issue in release binaries is separate, but this fixes the detection logic so that databases which already have FTS5 tables work correctly even after the build-tag fix is applied.

## Test plan

- [x] Verified on existing database with migration 3 already applied: `doctor` now reports `FTS5 true`
- [x] `messages search` uses FTS5 index (shows `[match]` highlighting instead of `Note: FTS5 not enabled`)
- [x] Clean database still works (migration 3 creates table and sets flag as before)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Media: recover panics per download job so one bad payload no longer drains the worker pool. (#179 — thanks @shaun0927)
 - Messages: show display text for replies, reactions, and media in `messages context`. (#183 — thanks @fuleinist)
+- Search: keep FTS5 enabled after reopening existing databases with already-applied migrations. (#185 — thanks @iamhitarth)
 - Send: persist retry-message plaintext so linked devices can decrypt retried messages. (#186 — thanks @SimDamDev)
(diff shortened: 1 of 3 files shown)


steipete:
Landed via rebase in 3dfd4f98a799a1a3a99d779165053d1fea751a55. Tightened the fix so existing empty FTS5 tables are detected on reopen, added the reopen regression, and ran full gate: `pnpm format:check && pnpm lint && pnpm test && pnpm build && git diff --check`. Thanks @iamhitarth.

PR #186, opened 2026-04-18
Enable UseRetryMessageStore so retry-receipts are handled

SimDamDev:
## Summary

When a recipient device cannot decrypt a message that wacli just sent (typically another linked device of the same account whose Signal session hasn't been freshly bootstrapped), it emits a retry-receipt. whatsmeow tries to find the original plaintext to re-encrypt it, but wacli doesn't persist outgoing messages, so the lookup fails with:

```
Failed to handle retry receipt for <jid>/<id> from <jid>:<n>: couldn't find message <id>
```

and the recipient is stuck on **"Waiting for this message. Check your phone."** indefinitely.

This PR sets `client.UseRetryMessageStore = true` right after `whatsmeow.NewClient`. whatsmeow then writes sent messages to the `retry_message_store` SQLite table, so a later wacli process (e.g. `wacli sync` restarting after a one-shot `wacli send`) can answer the retry-receipt and the recipient decrypts normally.

## How it breaks today

1. `wacli auth` — pair wacli to an account with at least one other linked device (WhatsApp Desktop / another linked device).
2. `wacli send text --to <yourself> --message "hi"` — the other linked devices fail to decrypt.
3. The sender sees the error above in logs; the recipient sees the "Waiting for this message" state forever (because `wacli send` has already exited and the in-memory retry cache died with it).

With this patch applied: retry-receipts are satisfied by the persistent store, message decrypts on every recipient device including ones that weren't online at send time.

## Scope

Six lines, one file (`internal/wa/client.go`). No new dependency; whatsmeow handles the `retry_message_store` table schema itself, so existing installs work without a migration.

## Test plan

- [x] Reproduce "Waiting for this message" on a linked recipient device with upstream v0.6.0
- [x] Rebuild wacli with this patch applied
- [x] Send the same message → recipient decrypts it correctly, even after the sending process has exited
- [x] `wacli sync` continues to work; no regression in inbound message handling

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.6.1 - Unreleased
 
+### Fixed
+
+- Send: persist retry-message plaintext so linked devices can decrypt retried messages. (#186 — thanks @SimDamDev)
+
 ### Docs
 
 - Maintainers: add CODEOWNERS and maintainer contact info.
--- a/internal/wa/client.go
+++ b/internal/wa/client.go
@@ -62,6 +62,12 @@ func (c *Client) init() error {
 
 	logger := waLog.Stdout("Client", "ERROR", true)
 	c.client = whatsmeow.NewClient(deviceStore, logger)
+	// Persist recently-sent messages so whatsmeow can answer retry-receipts
+	// across process restarts. Without this, recipients whose Signal session
+	// has not been freshly bootstrapped (typically other linked devices)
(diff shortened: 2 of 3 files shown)


steipete:
Landed via rebase onto main.

Added regression coverage for `UseRetryMessageStore` and fixed `internal/wa` to import the sqlite3 driver it directly depends on.

Thanks @SimDamDev!

Issue #189, opened 2026-04-21
Client outdated (405) connect failure on auth


stinct9:
**Describe the bug**
Running `wacli auth` fails with "Client outdated (405) connect failure". The QR channel closes immediately.

**To Reproduce**
1. Install wacli via `brew install steipete/tap/wacli` (version 0.2.0)
2. Run `wacli auth`
3. Error appears before QR code can be scanned

**Expected behavior**
QR code should display for authentication.

**Logs**
```
Starting authentication…
15:27:58.993[31m [Client ERROR] Client outdated (405) connect failure (client version: 2.3000.1030403648)[0m
QR channel closed
```

**Environment**
- wacli version: 0.2.0
- macOS: (Darwin 25.4.0 arm64)
- Install method: Homebrew

**Additional context**
WhatsApp appears to have updated their client version requirements. The hardcoded client version `2.3000.1030403648` is being rejected as outdated.

dinakars777:
Hey @stinct9 - This is now fixed in https://github.com/steipete/homebrew-tap/pull/25. You should be able to get 0.6.0 on Homebrew now, Please reopen if you still see any issues.

Issue #106, opened 2026-04-06
Client outdated (405) - Client version 2.3000.1030403648 blocked by WhatsApp


diogozsantospt:
**Problem**
WhatsApp is blocking the client version used by wacli. All commands fail with:

Client outdated (405) connect failure (client version: 2.3000.1030403648)
**Steps to reproduce**
1. Run any wacli command (e.g., `wacli auth` or `wacli send text`)
2. Observe the 405 error

**Environment**
- wacli version: 0.2.0
- Installed via: Homebrew (steipete/tap)
- OS: macOS (Apple Silicon)

**What I've tried**
- Reinstalling wacli: ❌ No effect
- Clearing session (`rm ~/.wacli/session.db`): ❌ No effect  
- Complete reinstall with clean session: ❌ No effect
**Root cause**
The underlying whatsmeow library needs to be updated to a newer version that uses a current WhatsApp client version.

**Request**
Please update the whatsmeow dependency to a newer version that supports the current WhatsApp client protocol.

Related: This appears to be a widespread issue when WhatsApp deprecates old client versions.

usamaejaz:
Same issue for me

architpai:
Hey even I was facing the same issue and got my claw to figure it out here is description of the fix for anybody who might find it useful:

The root cause is that wacli v0.2.0 embeds an outdated WhatsApp Web client protocol version. WhatsApp periodically deprecates old client versions, which breaks any app using a pinned old version.

The main branch already has updated whatsmeow dependencies — it just needs a new release.

Quick Fix (Build from Source)

Prerequisites:

• Go 1.24+ installed (go version)
• GCC/Clang installed (CGO required for SQLite support)

```
# 1. Clone the main branch (not the v0.2.0 release)
git clone --depth 1 https://github.com/steipete/wacli.git
cd wacli

# 2. Update dependencies to latest
go mod tidy

# 3. Build with CGO enabled (required for SQLite session storage)
go build -o wacli ./cmd/wacli

# 4. Install (optional — replaces old binary)
go install ./cmd/wacli

# 5. Verify
./wacli version    # Should show version > 0.2.0
macOS via Homebrew
```

If you originally installed via Homebrew, you'll need to build from source since the tap still points to v0.2.0:

```
# Clone and build as above, then replace Homebrew's binary:
git clone --depth 1 https://github.com/steipete/wacli.git ~/wacli-build
cd ~/wacli-build
go mod tidy
[...]

PR #191, opened 2026-04-21
ci: add Dependabot for Go modules and GitHub Actions

adhitShet:
Adds weekly grouped Dependabot updates for `gomod` and `github-actions`.

  This eliminates the recurring manual cycle where an old client version is deprecated, users start hitting 405 errors (#106, #189), and someone has to manually spot the whatsmeow fix and ship a release. Dependabot will open a PR within a week of any whatsmeow update — dinakars777 reviews, CI runs the gate, one-click merge.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,17 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


dinakars777:
Thanks for the suggestion! Dependabot is actually already enabled on this repo via GitHub's repository settings (see PR #69 as an example of it in action). Closing since this is already covered!

PR #195, opened 2026-04-28
fix: add warning log for dropped EncReactionMessage

### LINKED ISSUE #192: Reactions not appearing in messages list after interrupted sync
Reactions aren't showing up in `wacli messages list`, even after multiple syncs complete cleanly.

What happened: I sent a message via `wacli send text`. The recipient reacted with ❤️ - I got the native push notification. Ran `wacli messages list` right after syncing. Nothing.

**To reproduce:**
1. `wacli send text --to <JID> --message "..."`
2. Recipient reacts on their WhatsApp client
3. Native push notification confirms the reaction arrived on the device
4. Run `wacli sync` (default `--follow=true`) in the background, then stop it before it finishes
5. `wacli sync --once` -> wait for exit 0
6. `wacli sync --once` again
7. `wacli messages list --chat <JID> --limit 10`

The reaction is gone. Last entry in the list is still the original outgoing message - no `"Reacted ❤️ to ..."` row.

I dug into the source a bit. Encrypted reactions (`EncReactionMessage`) go through a decryption step in `sync_events.go`. My best guess: the reaction arrived during the first sync, which I killed (SIGTERM) mid-run. The two `--once` syncs after that didn't re-deliver the event. No warning, no error - it just silently disappeared.

Would be great if either missed events were re-fetched on reconnect, or at least a warning was logged when an `EncReactionMessage` is dropped so you know something was missed.

dinakars777:
Resolves #192. 

As identified by @matrixise and @fuleinist, if a sync is interrupted, incoming `EncReactionMessage` updates can fail to decrypt on subsequent syncs and get dropped silently. This PR adds a clear warning log to stderr when decryption fails, eliminating the silent failure and making it obvious to the user that an event was missed.

### DIFF
--- a/internal/app/sync_events.go
+++ b/internal/app/sync_events.go
@@ -58,7 +58,10 @@ func (a *App) addSyncEventHandler(ctx context.Context, opts SyncOptions, message
 func (a *App) handleLiveSyncMessage(ctx context.Context, opts SyncOptions, v *events.Message, messagesStored *atomic.Int64, enqueueMedia func(string, string)) {
 	pm := wa.ParseLiveMessage(v)
 	if pm.ReactionToID != "" && pm.ReactionEmoji == "" && v.Message != nil && v.Message.GetEncReactionMessage() != nil {
-		if reaction, err := a.wa.DecryptReaction(ctx, v); err == nil && reaction != nil {
+		reaction, err := a.wa.DecryptReaction(ctx, v)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "\rwarning: failed to decrypt reaction message %s: %v\n", pm.ID, err)
+		} else if reaction != nil {
 			pm.ReactionEmoji = reaction.GetText()
 			if pm.ReactionToID == "" {
 				if key := reaction.GetKey(); key != nil {


dinakars777:
Thanks @matrixise for the detailed repro steps — that makes debugging infinitely easier! Confirmed this is a real edge case. I just tested this locally by intentionally interrupting a --follow sync right as a reaction arrived, and it behaves exactly as you guessed: the WhatsApp server considers it delivered, our process dies before writing it to SQLite, and the event is never re-delivered on subsequent syncs.

@fuleinist — your intuition was spot on. I went ahead and opened PR #195 which adds a clear warning log to stderr when these decryption failures/drops happen.

Why just a warning for now? While wacli could try to automatically trigger a backfill when this happens, doing it reactively in the sync loop is risky. If decryption fails for a real cryptographic reason (e.g., missing keys), an auto-retry would trigger an infinite loop of fetching the same message, failing, and fetching again, which could get users rate-limited or banned by WhatsApp.

The "proper" architectural fix is to overhaul how wacli reconnects entirely: querying our SQLite DB for the last-seen-message-ID on startup and passing that to whatsmeow to force a backfill of anything missed during downtime. Since that touches core synchronization logic, I wanted to get the warning log merged first so users and automated AI agents aren't left in the dark failing silently while the larger fix is planned.

@matrixise, if you could test the warning log before it gets merged into main that would be helpful, you can compile the CLI directly from my PR branch using Go:

```bash
go install -tags sqlite_fts5 github.com/dinakars777/wacli/cmd/wacli@fix/dropped-reactions
```
Let me know if you're able to catch the new warning log in action!

steipete:
Thanks @dinakars777. I landed the same narrow warning behavior on main in f8ce9ee, with a regression test that proves encrypted reaction decrypt failures now emit stderr instead of failing silently. The changelog entry is in 1fb0170.\n\nLeaving #192 open because this does not claim to solve the larger interrupted-sync/backfill problem yet. Closing this PR as superseded by main.

PR #201, opened 2026-05-05
feat: add send text mentions

dinakars777:
## Summary

- add repeatable `--mention` to `wacli send text` for group messages
- normalize mention inputs via existing JID/phone parsing, strip device IDs, dedupe, and reject non-user mention targets
- document the new flag in README/docs and add focused regression coverage

## Issue

Addresses the mention support requested in #16. Reply-to support is already on `main`, so this PR is intentionally scoped to mentions only.

## Dependency checked

Current `go.mau.fi/whatsmeow@v0.0.0-20260427122815-7514259253a7` generated protobuf source exposes `ContextInfo.MentionedJID []string` and `GetMentionedJID()` in `proto/waE2E/WAWebProtobufsE2E.pb.go`. `send text` already uses `ExtendedTextMessage.ContextInfo` for replies/previews, so this patch sets that upstream field directly instead of relying on message text parsing.

## Tests

- `pnpm format:check`
- `go test ./cmd/wacli`
- `pnpm lint`
- `pnpm test`
- `pnpm build`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Auth: auto-detect a readable linked-device label and default linked-device platform to desktop. (#100 — thanks @pmatheus)
 - Profile: add `profile set-picture` to update the authenticated account profile picture from JPEG or PNG input. (#198 — thanks @gado-ships-it)
 - Send: add `send react` to add or clear reactions, with group sender validation. (#151 — thanks @draix)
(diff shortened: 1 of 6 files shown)


steipete:
Thanks @dinakars777. This is already landed on main in 09b2efb with docs/changelog and CI green, so this PR is now superseded by main. Closing to keep the queue clean.

PR #206, opened 2026-05-05
fix: interpret escape sequences (\n, \r, \t) in --message flag

slaveofcode:
Add unescapeMessage function to sendTextMessage so that --message "line1\nline2" properly renders as multi-line text on WhatsApp instead of showing literal \n characters.

### DIFF
--- a/cmd/wacli/auth.go
+++ b/cmd/wacli/auth.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


steipete:
Thanks @slaveofcode. I rebuilt this on main in de84bd2.

I kept the useful idea but made it opt-in as `send text --message-escapes` instead of changing all `--message` handling silently. That preserves existing scripts that intentionally send literal backslashes. I also left out the unrelated auth JSON QR change from this PR.

Proof: focused send tests, full local gate (`format:check`, `lint`, `test`, `build`, `git diff --check`), and a live WhatsApp send to the test chat where `\n` was stored/sent as an actual newline.

PR #225, opened 2026-05-08
feat(messages): add `download` command for media payloads

starshunter88:
## Summary

Adds a new subcommand:

```
wacli messages download --chat <jid> --id <msgid> [--output <path>]
```

Resolves the message in the local store, fetches the encrypted media blob from WhatsApp's CDN using the stored `direct_path` / `media_key` / hashes, and writes the decrypted bytes to disk via the existing `(*wa.Client).DownloadMediaToFile`.

When `--output` is omitted, the path defaults to the stored filename (sanitised) or `./<msgid>.<ext>` where the extension is derived from mime / media type.

## Why

`messages list / show / search` already surface media metadata, but there's currently no built-in way to pull the actual bytes. Useful for:

- Scripted audits where you want to attach a customer's WhatsApp screenshot to a ticket
- Personal CLI workflows where you want to grab an image / video / voice note without opening the WhatsApp UI
- Bug reproduction: triage a media report by hash without round-tripping through a phone

I bumped into this on a real customer-support thread today: needed to inspect 3 screenshots and a 32 MB video the customer sent, with `wacli` already running in a container — couldn't do it without leaving the CLI.

## Implementation

- New file `cmd/wacli/messages_download.go` with `newMessagesDownloadCmd`, plus three pure helpers: `defaultDownloadPath`, `sanitizeFilename`, `extensionFor`.
- Registered alongside the other `messages` subcommands in `cmd/wacli/messages.go`.
- Bails early with a clear error if the message has no `direct_path` / `media_key` / `file_enc_sha256` (i.e. not a media message, or the blob expired and was never re-fetched).
- Sanitises any user-supplied filename: only `[A-Za-z0-9._-]` survive, everything else becomes `_` so the path can't escape the working directory.
- Uses the existing `mediaType` field as `mmsType` — verbatim — same shape your encoders + decoders already produce upstream.

## Tests

`cmd/wacli/messages_download_test.go` covers the helpers (sanitisation, extension inference, default path). The actual download path goes through `(*wa.Client).DownloadMediaToFile` which already has fake coverage via `internal/app/fake_wa_test.go`.

```
$ go test -tags sqlite_fts5 -count=1 ./cmd/wacli/
ok  	github.com/openclaw/wacli/cmd/wacli	0.521s
```

## CHANGELOG

Added an `Unreleased` section. Happy to drop / reformat if you'd rather land it under a different version.

## Tested manually

- macOS local build: `go build -tags sqlite_fts5 -o /tmp/wacli ./cmd/wacli` → `wacli messages download --help` shows correctly.
- linux/amd64 (glibc) build via `golang:1.25-bookworm` → deployed in a Debian 12 container, downloaded 4 real WhatsApp media messages (3 images, 1 video, 32 MB) from a group chat, all decrypted correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- Messages: new `messages download` command — fetches the encrypted media payload of an image/video/audio/document/sticker message from WhatsApp's CDN and decrypts it to disk. Auto-derives the output filename from the stored filename or msg ID + mime/media-type extension when `--output` isn't given.
+
 ## 0.8.1 - 2026-05-08
 
 ### Changed
--- a/cmd/wacli/messages.go
+++ b/cmd/wacli/messages.go
@@ -26,6 +26,7 @@ func newMessagesCmd(flags *rootFlags) *cobra.Command {
 	cmd.AddCommand(newMessagesStarredCmd(flags))
 	cmd.AddCommand(newMessagesShowCmd(flags))
 	cmd.AddCommand(newMessagesContextCmd(flags))
+	cmd.AddCommand(newMessagesDownloadCmd(flags))
 	cmd.AddCommand(newMessagesExportCmd(flags))
 	cmd.AddCommand(newMessagesDeleteCmd(flags))
 	cmd.AddCommand(newMessagesEditCmd(flags))
--- a/cmd/wacli/messages_download.go
+++ b/cmd/wacli/messages_download.go
@@ -1,0 +1,182 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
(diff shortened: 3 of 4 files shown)


dinakars777:
Thanks for the PR. I think this overlaps with existing functionality on current `main`: `wacli media download --chat JID --id MSG_ID [--output PATH]`, documented in `docs/media.md`.

As written, this adds a second download path under `messages download` that duplicates the existing media command and bypasses some existing behavior, including `ResolveMediaOutputPath` and marking `local_path` / `downloaded_at`.

I’m going to close this version as superseded by the existing `media download` command. If you want to pursue the ergonomics angle, a smaller follow-up that makes `messages download` delegate to the existing media download implementation would be easier to review.

PR #13, opened 2026-01-27
feat: show full message IDs when piped or with --full flag

rickhallett:
Message IDs in tabular output (list/search/context) were truncated to 14 chars, making them unusable with `messages show --id` or `media download --id` without resorting to `--json | jq`.

Changes:
- Add `--full` flag to disable truncation in table output
- Auto-detect piped output and show full IDs (uses existing `isTTY()` helper)
- Add tests for helper functions

 - Closes the workflow gap where users couldn't copy IDs from list output.

PR #226, opened 2026-05-08
feat: preserve interactive message buttons end-to-end

### LINKED ISSUE #223: `messages list` JSON output drops buttons and list options from interactive messages
## Summary

  `messages list --json` silently discards interactive elements from several WhatsApp Business message types. The
  `Text`/`DisplayText` fields are populated correctly, but any buttons, URL actions, or list rows that appear in the WhatsApp UI
  are not stored or returned.

  ## Affected message types

  | Type | What's missing |
  |---|---|
  | `TemplateMessage` (HydratedFourRowTemplate) | URL buttons, quick-reply buttons, call buttons from `GetHydratedButtons()` |
  | `ButtonsMessage` | Quick-reply buttons from `GetButtons()` |
  | `ListMessage` | The "Options" button label (`GetButtonText()`) and all selectable rows from `GetSections()` |

  ## Root cause

  `extractBusinessText()` in `internal/wa/messages_business.go` only reads prose text fields (title, body, footer) from each
  message type and ignores the interactive payload entirely. The `Message` struct and database schema have no `buttons` field, so
   the data has nowhere to go.

  ## Reproduction

  ```sh
  wacli messages list --chat "<business_number>" --from-them --json --limit 5 | jq '.[].Buttons'
  # null for all messages, even those with visible buttons in the WhatsApp UI
  ```

  ## Proposed fix

  We have a working implementation that:

  1. Adds a `Button` struct (`type`, `display_text`, `id`, `url`, `phone_number`, `description`)
  2. Populates `Buttons []Button` in `ParsedMessage` for all three affected message types
  3. Stores buttons as a JSON column (`buttons TEXT`) via a new schema migration
  4. Returns `Buttons` in all JSON output (omitted when empty, so existing responses are unaffected)
  5. Also fixes `migrateLIDMessagesToPN` in `lid_migration.go`, which had the same hardcoded column list and would silently drop
  buttons during LID→PN migration

  Happy to open a PR if that's welcome.

ignaciovarela:
⏺ Closes #223

  ## Summary

  `messages list --json` silently discarded interactive elements from WhatsApp Business message types. Text and display text were captured correctly, but buttons, URL actions, and list rows visible in the WhatsApp UI were dropped at the parser and never stored.

  This PR captures and stores them across the full pipeline: parser → store migration → JSON output. The `Buttons` field is omitted when empty so existing JSON responses are unaffected.

  ## Changes

  **Parser (`internal/wa`)**
  - Add `Button` struct: `type`, `display_text`, `id`, `url`, `phone_number`, `description`
  - Add `Buttons []Button` to `ParsedMessage`
  - `extractBusinessText` now populates buttons independently of the text guard for:
    - `TemplateMessage` (HydratedFourRowTemplate): `url`, `quick_reply`, `call` buttons via `GetHydratedButtons()`
    - `ButtonsMessage`: `quick_reply` buttons with ID via `GetButtons()`
    - `ListMessage`: list opener label via `GetButtonText()` + one `list_row` per row in `GetSections()`

  **Store (`internal/store`)**
  - Add `Button` struct and `Buttons []Button` to `Message`
  - Add `buttons TEXT` column to schema (migration #13, with table-existence guard for partial-schema test databases)
  - Serialize/deserialize buttons as JSON in `UpsertMessage`, `scanMessages`, and `GetMessage`
  - Fix `migrateLIDMessagesToPN`: add `buttons` to its hardcoded column list so interactive data is not dropped during LID→PN chat migration

  **Wire-up (`internal/app`)**
  - `waButtonsToStore` helper converts `[]wa.Button` → `[]store.Button` in `sync.go`

  ## Tests

  - `TestParseTemplateMessageWithURLButtons`: two URL buttons, display text and URL captured
  - `TestParseTemplateMessageWithMixedButtons`: quick-reply + call buttons
  - `TestParseButtonsMessage`: updated to assert button IDs and display text
  - `TestParseListMessage`: updated to assert list opener button + two list rows with ID and description

  ## Local testing

Built locally with `CGO_ENABLED=1 go build -tags sqlite_fts5 -o dist/wacli ./cmd/wacli` and verified against real WhatsApp Business messages

### DIFF
--- a/internal/app/sync.go
+++ b/internal/app/sync.go
@@ -393,6 +393,7 @@ func (a *App) storeParsedMessage(ctx context.Context, pm wa.ParsedMessage) error
 		FromMe:          pm.FromMe,
 		Text:            pm.Text,
 		DisplayText:     displayText,
+		Buttons:         waButtonsToStore(pm.Buttons),
 		IsForwarded:     pm.IsForwarded,
 		ForwardingScore: pm.ForwardingScore,
 		ReactionToID:    pm.ReactionToID,
@@ -421,6 +422,24 @@ func (a *App) storeParsedMessage(ctx context.Context, pm wa.ParsedMessage) error
 		})
 	}
 	return nil
+}
+
+func waButtonsToStore(buttons []wa.Button) []store.Button {
+	if len(buttons) == 0 {
+		return nil
+	}
+	out := make([]store.Button, len(buttons))
+	for i, b := range buttons {
+		out[i] = store.Button{
+			Type:        b.Type,
+			DisplayText: b.DisplayText,
+			ID:          b.ID,
+			URL:         b.URL,
+			PhoneNumber: b.PhoneNumber,
+			Description: b.Description,
+		}
+	}
+	return out
 }
 
 func (a *App) buildDisplayText(ctx context.Context, pm wa.ParsedMessage) string {
--- a/internal/store/lid_migration.go
+++ b/internal/store/lid_migration.go
@@ -151,7 +151,7 @@ func migrateLIDMessagesToPN(tx *sql.Tx, lidJID, pnJID string) error {
 			is_forwarded, forwarding_score, reaction_to_id, reaction_emoji,
 			media_type, media_caption, filename, mime_type, direct_path,
 			media_key, file_sha256, file_enc_sha256, file_length, local_path, downloaded_at,
-			revoked, deleted_for_me
+			revoked, deleted_for_me, buttons
 		)
 		SELECT
 			?,
@@ -179,7 +179,8 @@ func migrateLIDMessagesToPN(tx *sql.Tx, lidJID, pnJID string) error {
 			local_path,
 			downloaded_at,
 			revoked,
-			deleted_for_me
+			deleted_for_me,
+			buttons
 		FROM messages
 		WHERE chat_jid = ?
 		ON CONFLICT(chat_jid, msg_id) DO UPDATE SET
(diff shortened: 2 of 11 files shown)


dinakars777:
Thanks, this is a good direction and the parser coverage looks useful.

Before merging, could you please run gofmt and add store-level regression coverage for the new `buttons` field? The PR currently fails `pnpm format:check` on `internal/store/types.go`.

One behavioral issue to fix: now that buttons are persisted message content, the scrub paths should clear them too. `MarkMessageRevoked`, `MarkMessageDeletedForMe`, and likely `UpdateMessageText` clear text/media fields but leave `buttons` intact, so stale interactive data can remain visible through direct message reads after revoke/delete/edit. Please set `buttons = NULL` in those paths and cover that with tests.

ignaciovarela:
> Thanks, this is a good direction and the parser coverage looks useful.
> 
> Before merging, could you please run gofmt and add store-level regression coverage for the new `buttons` field? The PR currently fails `pnpm format:check` on `internal/store/types.go`.
> 
> One behavioral issue to fix: now that buttons are persisted message content, the scrub paths should clear them too. `MarkMessageRevoked`, `MarkMessageDeletedForMe`, and likely `UpdateMessageText` clear text/media fields but leave `buttons` intact, so stale interactive data can remain visible through direct message reads after revoke/delete/edit. Please set `buttons = NULL` in those paths and cover that with tests.

Addressed all three points in three separate commits:

  - style: fixed the gofmt alignment issue in store.Message
  - fix: added buttons = NULL to MarkMessageRevoked, MarkMessageDeletedForMe, and UpdateMessageText so stale interactive data is cleared along with text and media
  - test: added five store-level regression tests covering buttons round-trip through GetMessage and ListMessages, and clearing on each of the three scrub paths

dinakars777:
Thanks for addressing the earlier feedback.

One final ask before merge: since this PR explicitly updates `migrateLIDMessagesToPN`, could you add a narrow store-level regression test proving `MigrateLIDToPN` preserves `buttons`?

A good test would insert a message with buttons under a `@lid` chat, run `MigrateLIDToPN(lid, pn)`, then assert the migrated PN message still has the same button metadata. That gives us coverage for the hardcoded migration column list this PR updates.

ignaciovarela:
Added `TestMigrateLIDToPNPreservesButtons` to `internal/store/lid_migration_test.go`. It inserts a message with buttons under a @lid chat, runs MigrateLIDToPN, and asserts the migrated PN message carries the same button slice — directly covering the column list change in migrateLIDMessagesToPN.

Thanks for the thorough review, really appreciated the feedback. Let me know if there's anything else you'd like me to address before merge!

dinakars777:
Thanks @ignaciovarela for the quick follow-ups and for adding the targeted regression coverage. This is merged now.

Issue #14, opened 2026-01-28
Feature request: Support for WhatsApp voice messages (PTT)


shubs:
## Summary

Currently `wacli send file` sends audio files as documents/attachments, not as native WhatsApp voice messages (Push-To-Talk/PTT).

## Current Behavior

When sending an audio file (e.g., `.ogg`, `.opus`, `.mp3`), it appears as a downloadable attachment in the chat, not as the inline voice message with waveform that WhatsApp users expect.

```bash
wacli send file --to "+1234567890" --file voice.ogg --mime "audio/ogg; codecs=opus"
# → Sends as document attachment, not as voice note
```

## Desired Behavior

Add a `--ptt` or `--voice` flag to send audio as a proper WhatsApp voice message:

```bash
wacli send file --to "+1234567890" --file voice.ogg --ptt
# → Sends as native voice message with waveform
```

## Use Case

This is particularly useful for bots and automation that need to send voice messages (e.g., TTS responses, voice notifications) in a natural way that recipients expect from WhatsApp.

## Technical Notes

WhatsApp PTT messages use OGG Opus format with specific metadata flags in the protobuf message. The `whatsmeow` library likely supports this via the `AudioMessage` type with `PTT: true`.

dinakars777:
Closing this one to track in https://github.com/steipete/wacli/issues/40, please reopen if this is still needed. Thanks for your time.

PR #230, opened 2026-05-09
feat: implement WhatsApp polls (send/vote/decrypt/query)

Ortes:
## Summary
Wires the whatsmeow poll API end-to-end through the existing wacli layers: send polls (single + multi-select), persist incoming polls (V1/V2/V3/V4-FutureProof/V5 variants), vote on polls, decrypt inbound vote events, and query results with per-voter breakdown. Storage: two new tables (`polls`, `poll_votes`) introduced via migration #14.

## What's new
- `wacli send poll --to … --question … --option … [--multi N] [--ephemeral]`
- `wacli poll vote --to … --id … --option …`
- `wacli poll show --to … --id …` (aggregates + per-voter list, `--json` supported)
- `wacli polls list [--chat … --limit N]`
- IPC delegate paths for `poll` and `poll_vote` so the commands work while `wacli sync` holds the lock.
- New `wa.Client` methods: `SendPoll`, `SendPollVote`, `DecryptPollVote`, plus `WAClient` interface extension.

## LID-related subtleties caught while testing live
1. Self-poll vote events arrive keyed under the LID rather than the phone JID the poll was stored under. Falling back to a msg-id lookup (`store.FindPollByMsgID`) is required for the live decrypt path.
2. Voter JIDs are normalised via `ResolveLIDToPN` before upsert so the outbound vote row and the inbound echo collapse into a single `poll_votes` row.
3. On accounts with `LIDMigrationTimestamp > 0`, whatsmeow's `SendMessage` rewrites the destination from phone JID to LID. We pre-translate `pollInfo.Chat`/`Sender` in `Client.SendPollVote` so the embedded `PollCreationMessageKey` matches the wire destination — otherwise the recipient cannot link the vote to the poll it references and the vote silently no-ops on the mobile side.
4. `MessageInfo.IsFromMe` set from `messages.from_me` so `PollCreationMessageKey.FromMe` is correct when voting on our own polls.

## Test plan
- [x] `go test ./...` passes (all 13 packages).
- [x] Live: send single-select poll → renders on mobile.
- [x] Live: send multi-select poll → mobile shows "Select up to N".
- [x] Live: vote from mobile → sync ingests, `poll show` reflects selection.
- [x] Live: vote from CLI (`wacli poll vote`) → mobile shows the vote (after the LID translation fix).
- [x] Live: change vote → upsert correctly replaces, single voter row remains.
- [ ] Live: vote from a second account on a third party's poll (not exercised; covered by unit tests).
- [ ] Live: history sync of polls + votes (not exercised live; code path mirrors the live path).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Docker: add a local image with `/data` persistence, bundled `ffmpeg`, and Docker CI smoke coverage.
+- Polls: add sending, voting, local result inspection, and sync persistence for WhatsApp polls. (#230 - thanks @Ortes)
 - Send: add opt-in `send text --ephemeral` wrapping for disappearing-message chats. (#227 - thanks @AndroidPoet)
 
 ### Security
--- a/cmd/wacli/poll_cmd.go
+++ b/cmd/wacli/poll_cmd.go
@@ -262,6 +262,7 @@ func buildPollVoteInfo(a *app.App, chat types.JID, pollMsgID, senderOverride str
 			Chat:     chat,
 			Sender:   senderJID,
 			IsFromMe: fromMe,
+			IsGroup:  chat.Server == types.GroupServer,
 		},
 		ID:        pollMsgID,
 		Timestamp: ts,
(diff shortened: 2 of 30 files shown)


steipete:
Triage note: leaving this open, but it needs a maintainer pass before it can land.

Current state from GitHub is `mergeStateStatus: DIRTY`, so the branch needs to be rebased or updated against current main first. The patch is also a large feature surface: new poll commands, storage migrations, sync handling, IPC paths, and whatsmeow client methods. Next useful step is to update the branch, then run the full local gate and review the migration/API surface carefully.

PR #232, opened 2026-05-11
feat: extract buttons from InteractiveMessage NativeFlowMessage

ignaciovarela:
NativeFlowMessage is the newer WhatsApp interactive type used for CTA
  URL, quick-reply, and call buttons. Two fixes:

  1. The pm.Text == "" guard on the InteractiveMessage block was skipping button extraction entirely whenever plain text was already set. Text and button extraction are now independent.

  2. Add nativeFlowButton() to parse NativeFlowMessage buttons by decoding ButtonParamsJSON for cta_url, quick_reply, and cta_call button name values.

### DIFF
--- a/internal/wa/messages_business.go
+++ b/internal/wa/messages_business.go
@@ -1,6 +1,7 @@
 package wa
 
 import (
+	"encoding/json"
 	"strings"
 
 	waProto "go.mau.fi/whatsmeow/binary/proto"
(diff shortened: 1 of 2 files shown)


steipete:
Thanks @ignaciovarela. This is fixed on main in openclaw/wacli@cac5f20, with NativeFlow parsing and regression tests for the interactive-message/template paths.

I am closing this draft PR because the equivalent fix has already landed on main, and the changelog credits both related contributors for #233.

PR #234, opened 2026-05-12
fix: resolve privacy token failure for new contacts by warming up recipient before send

### LINKED ISSUE #212: Messages not being sent even though authenticated.
I have my WaCli Authenticated via login, but still it isnt sending messages. It says send successfully after throwing a error but the message does not get sent. 

Image

Image

Image

AndroidPoet:
## Summary

Fixes messages silently not being delivered to new/unknown contacts, despite `send text` returning `sent:true`.

## Problem

When sending a message to a recipient with no prior conversation state, whatsmeow's privacy token (tctoken) IQ fails with `400: bad-request`. The message is sent without the `tctoken` node and WhatsApp silently drops it.

Stderr shows:
```
Failed to issue privacy token for 55XXXXXXXXXXX@s.whatsapp.net: info query returned status 400: bad-request
```

## Root cause

The privacy token IQ fails because WhatsApp's server has no prior state (LID mapping, device list) for the recipient. Users who already have an existing chat/contact history do not encounter this issue.

## Fix

Call `GetUserInfo` (which issues a `usync` query) on the recipient before sending to establish the required server-side state. After the usync resolves the recipient's devices and LID mapping, the subsequent privacy token issuance succeeds.

Key design:
- **Best-effort**: failures are logged to stderr but never block the send
- **Scoped to user JIDs**: skips groups, newsletters, broadcast lists
- **All send commands covered**: `text`, `file`, `sticker`, `voice`, `react`

## Testing

- Build: `pnpm build` — passes
- Lint: `pnpm lint` — passes
- Format: `pnpm format:check` — passes
- All tests: `pnpm test` — passes (3 new regression tests added)

Closes #212

### DIFF
--- a/cmd/wacli/send.go
+++ b/cmd/wacli/send.go
@@ -105,6 +105,7 @@ func newSendTextCmd(flags *rootFlags) *cobra.Command {
 			if err := a.Connect(ctx, false, nil); err != nil {
 				return err
 			}
+			warmupRecipient(ctx, a.WA(), toJID, os.Stderr)
 			if err := warnRapidSendIfNeeded(a.StoreDir(), time.Now().UTC(), os.Stderr); err != nil {
 				return err
 			}
--- a/cmd/wacli/send_file_cmd.go
+++ b/cmd/wacli/send_file_cmd.go
@@ -77,6 +77,7 @@ func newSendFileCmd(flags *rootFlags) *cobra.Command {
 			if err := a.Connect(ctx, false, nil); err != nil {
 				return err
 			}
+			warmupRecipient(ctx, a.WA(), toJID, os.Stderr)
 			if err := warnRapidSendIfNeeded(a.StoreDir(), time.Now().UTC(), os.Stderr); err != nil {
 				return err
 			}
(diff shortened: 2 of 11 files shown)


dinakars777:
Thanks for chasing this down. The direction makes sense, and the direct send paths look covered.

I don’t think this is ready to merge yet because the lock-delegated send paths still bypass the warm-up. When `sync` owns the store lock, `tryDelegateSend` routes through `executeDelegatedText`, `executeDelegatedFile`, `executeDelegatedSticker`, and `executeDelegatedReact` in `cmd/wacli/send_ipc.go`; those paths still go straight to `warnRapidSendIfNeeded` / `runSendOperation` without calling `warmupRecipient`.

Can you add the same best-effort warm-up to the delegated paths, with a small regression test if practical? Otherwise this only fixes direct sends and should not `Closes #212` yet.

AndroidPoet:
Thanks for flagging this you were absolutely right about the delegated path gap. I’ve now wired the same best-effort warm-up into the lock-delegated send handlers in cmd/wacli/send_ipc.go (text, file, sticker, and react), so those flows no longer skip warm-up when sync holds the store lock. Commit is e2b06c6 on fix/privacy-token-warmup, and go test ./cmd/wacli is passing.

dinakars777:
Thanks @AndroidPoet , I am keeping #212 open for now.

ceifa:
Thanks for making this fix!! Looking forward to test on the next release

PR #237, opened 2026-05-13
fix(wa): extract NativeFlowMessage buttons from InteractiveMessage

### LINKED ISSUE #233: InteractiveMessage with NativeFlowMessage buttons not extracted (CTA URL, quick-reply, call)
## Problem

Following up on #226 — buttons from `InteractiveMessage` with a `NativeFlowMessage` payload are not extracted. 

This is the message type used by the [WhatsApp Business Platform for interactive CTA URL buttons](https://developers.facebook.com/docs/whatsapp/cloud-api/messages/interactive-cta-url-messages) — e.g. "Track your order", "Complete payment", "Book appointment" — as well as quick-reply and  call buttons sent through the same format.

  Current `wacli messages list --json` output:
  ```json
  {
    "Text": "Your order has shipped. Track it using the link below.",
    "Buttons": null
  }
  ```

  Expected:
  ```json
  {
    "Text": "Your order has shipped. Track it using the link below.",
    "Buttons": [
      { "type": "url", "display_text": "Track order", "url": "https://..." }
    ]
  }
  ```

  ## Root cause

  Two issues in `internal/wa/messages_business.go`:

  1. **`NativeFlowMessage` not handled at all.** Each button carries a `Name` field (e.g. `cta_url`,
     `quick_reply`, `cta_call`) and a `ButtonParamsJSON` string that encodes the button parameters
     (`display_text`, `url`, `id`, `phone_number`).

  2. **`pm.Text == ""` guard blocks button extraction.** The `InteractiveMessage` block is currently
     gated on `if im != nil && pm.Text == ""`. Since these messages always carry body text, the
     condition is never true and button extraction is skipped entirely.

## Feedback needed

I have a working version in this [draft PR](https://github.com/openclaw/wacli/pull/232), feedback on the approach before marking this ready for review would be much appreciated.

mturac:
Fixes #233

## Problem
`InteractiveMessage` payloads carrying a `NativeFlowMessage` block (CTA URL, quick-reply, call buttons used by the WhatsApp Business Cloud API) were dropped: `Buttons` came back `null` even when the upstream message carried them.

## Root cause (verified against `internal/wa/messages_business.go`)

1. `NativeFlowMessage` was not handled in `extractBusinessText`. Each button there carries a `Name` field (`cta_url`, `quick_reply`, `cta_call`) plus a `ButtonParamsJSON` payload, neither of which were being parsed.
2. The interactive branch was gated by `pm.Text == ""`, which is virtually never true for these messages — so button extraction was being short-circuited.

## Fix

- Drop the `pm.Text == ""` guard around the `InteractiveMessage` branch so button extraction runs unconditionally.
- Parse `ButtonParamsJSON` and append entries to `pm.Buttons` based on `Name`:
  - `cta_url` → `Button{Type:"url", DisplayText, URL}`
  - `quick_reply` → `Button{Type:"quick_reply", DisplayText, ID}`
  - `cta_call` → `Button{Type:"call", DisplayText, PhoneNumber}`

The extracted `Text` field still falls back to `interactiveText(im)` when empty.

## Testing

- `go build ./...` — clean
- `go test ./internal/wa/...` — passes
- `go vet ./...` — clean

Happy to iterate on review feedback.

### DIFF
--- a/internal/wa/messages_business.go
+++ b/internal/wa/messages_business.go
@@ -1,6 +1,7 @@
 package wa
 
 import (
+	"encoding/json"
 	"strings"
 
 	waProto "go.mau.fi/whatsmeow/binary/proto"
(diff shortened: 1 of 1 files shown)


steipete:
Thanks @mturac. This is fixed on main in openclaw/wacli@cac5f20, with NativeFlow parsing and regression tests for the interactive-message/template paths.

I am closing this PR because the equivalent fix has already landed on main, and the changelog credits both related contributors for #233.

PR #238, opened 2026-05-13
fix: include disappearing context for ephemeral text

dovocoder:
## Summary
- add disappearing-message context metadata when `send text --ephemeral` is used
- send plain ephemeral text as extended text so the context can be attached
- cover both plain and link-preview ephemeral text in tests
- update send docs to mention the extra disappearing context

## Test Plan
- `gofmt -l .`
- `go vet ./...`
- `go test ./...`
- `go test -tags sqlite_fts5 ./...`
- `GOOS=windows GOARCH=amd64 go test -c ./internal/lock -o /tmp/wacli-lock-windows.test.exe`
- `CGO_ENABLED=0 go build ./cmd/wacli` fails with the expected CGO-required sentinel
- `CGO_ENABLED=1 CGO_CFLAGS="-Wno-error=missing-braces" go build -tags sqlite_fts5 -o dist/wacli ./cmd/wacli`
- `git diff --check`

### DIFF
--- a/cmd/wacli/send.go
+++ b/cmd/wacli/send.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -42,6 +43,7 @@ func newSendTextCmd(flags *rootFlags) *cobra.Command {
 	var replyToSender string
 	var noPreview bool
 	var ephemeral bool
+	var ephemeralDuration string
 	var messageEscapes bool
 	postSendWait := postSendRetryReceiptWait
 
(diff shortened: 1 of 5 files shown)


dinakars777:
Thanks for the fix and the tests. I don’t think this is quite ready yet because the new expiration lookup runs for every text send, not only ephemeral sends.

`sendTextMessage` now calls `ephemeralExpiration(ctx, a.WA(), to)` before `sendTextMessageWithSender`, so a normal non-ephemeral group send can still call `GetGroupInfo` even though it will take the plain `SendText` path and never use the expiration. That adds an avoidable network call and a new failure/latency surface to ordinary sends.

Can you move the expiration lookup so it only runs when `--ephemeral` is enabled? Ideally the non-ephemeral plain-text fast path should stay exactly as cheap as before.

dovocoder:
Hi, sure I will fix it 👍

dinakars777:
Thanks @dovocoder!

PR #240, opened 2026-05-15
Fix send target canonicalization for phone-number recipients

### LINKED ISSUE #212: Messages not being sent even though authenticated.

ceifa:
## Summary
- canonicalize direct phone-number send targets through WhatsApp registration lookup before sending
- keep the existing GetUserInfo warmup, but run it against the canonical JID returned by IsOnWhatsApp
- add coverage for BR-style phone input where the raw phone JID differs from the canonical WhatsApp JID

## Problem
In #212, direct sends could return `sent:true` while logging:

```
Failed to issue privacy token ... 400: bad-request
```

In my repro, this happened for Brazilian mobile numbers where the input had the extra ninth digit. For example, the raw input shape `+55XX9XXXXXXXX` was sent as `55XX9XXXXXXXX@s.whatsapp.net`, but WhatsApp registration lookup canonicalized it to `55XXXXXXXXXX@s.whatsapp.net`.

Sending to the raw JID meant `GetUserInfo` returned no useful LID/device state, the trusted-contact token request failed, and the message was not visible/delivered despite stdout reporting success.

## Fix
The send warmup now:
1. detects direct PN recipients
2. calls `IsOnWhatsApp("+" + jid.User)`
3. replaces the destination with the returned canonical JID when the number is registered
4. runs `GetUserInfo` against that canonical JID before sending

## Validation
- `go test ./...`
- local live retest against the previously failing direct BR recipients: sends used canonical JIDs and no longer emitted the privacy-token 400

Fixes #212

## Human note

I don't know golang, I asked my claw to try fix it and it did, then I asked to open a PR and it probably has slop. Feel free to close but you can use it as a reference to actually fix the problem.

### DIFF
--- a/cmd/wacli/send.go
+++ b/cmd/wacli/send.go
@@ -105,7 +105,7 @@ func newSendTextCmd(flags *rootFlags) *cobra.Command {
 			if err := a.Connect(ctx, false, nil); err != nil {
 				return err
 			}
-			warmupRecipient(ctx, a.WA(), toJID, os.Stderr)
+			toJID = warmupRecipient(ctx, a.WA(), toJID, os.Stderr)
 			if err := warnRapidSendIfNeeded(a.StoreDir(), time.Now().UTC(), os.Stderr); err != nil {
 				return err
 			}
--- a/cmd/wacli/send_helpers.go
+++ b/cmd/wacli/send_helpers.go
@@ -131,10 +131,10 @@ func warnRapidSendIfNeeded(storeDir string, now time.Time, stderr io.Writer) err
 	return nil
 }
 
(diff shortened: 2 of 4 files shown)


steipete:
Thanks @ceifa. The fix is landed on main in openclaw/wacli@2ab69c3.

Main now canonicalizes direct phone-number recipients through `IsOnWhatsApp` before warmup/send, includes regression coverage for the regional rewrite case, and has the changelog entry crediting #212/#240. I am closing this PR because the equivalent fix has already landed.

PR #242, opened 2026-05-15
Persist WhatsApp call events

### LINKED ISSUE #231: Persist WhatsApp call events / call log metadata instead of storing them as "(message)"
## Summary

`wacli sync` currently appears to drop or flatten WhatsApp call-related events. In the local message store, a recent WhatsApp call/system event is persisted as a normal message with `display_text = "(message)"` and no structured metadata, which makes it impossible for downstream tools to detect that a conversation was resolved by a later WhatsApp call.

## Observed behavior

After syncing a chat where a WhatsApp call occurred, the relevant row in `wacli.db.messages` looks like this:

```sql
SELECT rowid, chat_jid, chat_name, msg_id, sender_jid, sender_name, ts, from_me,
       text, media_type, display_text
FROM messages
WHERE chat_jid = '166752138866699@lid'
ORDER BY ts DESC
LIMIT 5;
```

Representative result:

```text
chat_name:    Jon - ...
chat_jid:     166752138866699@lid
from_me:      0
text:         NULL
media_type:   NULL
display_text: (message)
```

There is no call direction, call state/outcome, duration, call ID, media type, or participant metadata available in the durable `messages` table.

## Why this matters

Tools like loopclose/response trackers need to mark a waiting thread as resolved when the user later calls the person on WhatsApp. This is currently possible for iMessage/Phone because macOS exposes CallHistory, but WhatsApp calls are invisible after `wacli` reduces the event to `(message)`.

Desired downstream rule:

> If a WhatsApp thread is waiting on the user and there is a later connected/outbound WhatsApp call with the same chat/contact, mark the thread resolved.

That rule needs durable call metadata from `wacli`.

## Source notes

The current sync path seems to be:

- `internal/app/sync_events.go`: `addSyncEventHandler` handles `*events.Message` via `handleLiveSyncMessage`
- `internal/wa/messages.go`: `ParseLiveMessage` reduces `events.Message` into `ParsedMessage`
- `internal/app/sync.go`: `storeParsedMessage` writes `ParsedMessage` into `messages`

whatsmeow appears to expose call-related event types separately in `go.mau.fi/whatsmeow/types/events/call.go`, including:

- `CallOffer`
- `CallAccept`
- `CallPreAccept`
- `CallTransport`
- `CallOfferNotice`
- `CallRelayLatency`
- `CallTerminate`
- `CallReject`
- `UnknownCallEvent`

whatsmeow also has proto types for historical call log messages, including `CallLogMessage` and outcomes such as `CONNECTED`, `MISSED`, `FAILED`, `REJECTED`, and `ACCEPTED_ELSEWHERE`.

## Proposed implementation

Add durable call-event persistence.

Possible schema:

```sql
CREATE TABLE call_events (
  rowid INTEGER PRIMARY KEY AUTOINCREMENT,
  chat_jid TEXT NOT NULL,
  sender_jid TEXT,
  call_id TEXT,
  event_type TEXT NOT NULL,      -- offer, accept, terminate, reject, call_log, unknown
  direction TEXT,                -- inbound/outbound/unknown, best-effort
  media TEXT,                    -- audio/video/unknown, best-effort
  outcome TEXT,                  -- connected/missed/failed/rejected/etc when available
  reason TEXT,
  ts INTEGER NOT NULL,
  raw_json TEXT,                 -- optional/debug, omit or redact if too sensitive
  UNIQUE(chat_jid, call_id, event_type, ts)
);
CREATE INDEX idx_call_events_chat_ts ON call_events(chat_jid, ts);
CREATE INDEX idx_call_events_ts ON call_events(ts);
```

Then expose it through one of:

- `wacli calls list [--chat <jid>] [--json]`
- include call events in `messages list --json` with a distinct `message_type` / `system_type`
- both, if backwards-compatible

## Acceptance criteria

- Live call events emitted by whatsmeow are not silently ignored.
- At least call ID, chat JID, event type, timestamp, and best-effort direction are persisted.
- If call log proto messages are received through history sync, they are parsed instead of becoming `(message)`.
- `wacli calls list --json`, or equivalent, can show recent call events.
- Existing message listing behavior remains backwards-compatible.
- Add tests for one live call event and one call-log/proto message path if fixtures can be built.

## Notes

This would let downstream tools safely close “waiting on user” threads only after a later connected/outbound WhatsApp call, without guessing from a generic `(message)` row.

dovocoder:
Fixes #231

## Summary
- persist WhatsApp call metadata in a new `call_events` table instead of flattening call-log messages to generic `(message)` rows
- parse live whatsmeow call events and historical `CallLogMessage` payloads into structured call events
- add `wacli calls list` with table and `--json` output, plus docs and changelog coverage
- keep existing message listing behavior backwards-compatible while exposing calls through the dedicated command

## Details
- stores chat/sender names, call ID, message ID, event type, direction, media, outcome, reason, call type, duration, timestamp, and participants where available
- adds migration 15 for `call_events`; upstream polls remain migration 14 after the rebase
- handles app-state call logs and message-carried call logs, plus live call offer/accept/terminate/reject-style events
- adds tests for call-event upsert/listing, schema repair, message parsing, sync handling, and CLI output

## Validation
- `env GOCACHE=/tmp/go-build go test ./internal/store ./internal/wa`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.9.1 - Unreleased
 
+### Added
+
+- Calls: persist WhatsApp call signaling and call-log metadata, and add `wacli calls list`.
+
 ## 0.9.0 - 2026-05-15
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ More recipes — replies, mentions, stickers, voice, reactions, channels, histor
 | Area | Pages |
 | --- | --- |
 | **Setup** | [overview](docs/overview.md) · [auth](docs/auth.md) · [accounts](docs/accounts.md) · [sync](docs/sync.md) · [doctor](docs/doctor.md) |
-| **Messaging** | [messages](docs/messages.md) · [send](docs/send.md) · [media](docs/media.md) · [presence](docs/presence.md) |
+| **Messaging** | [messages](docs/messages.md) · [calls](docs/calls.md) · [send](docs/send.md) · [media](docs/media.md) · [presence](docs/presence.md) |
 | **Address book** | [contacts](docs/contacts.md) · [chats](docs/chats.md) · [groups](docs/groups.md) · [channels](docs/channels.md) |
 | **History** | [history coverage / fill / backfill](docs/history.md) |
 | **Local store** | [store](docs/store.md) · [companion integrations](docs/integrations.md) |
--- a/cmd/wacli/calls.go
+++ b/cmd/wacli/calls.go
@@ -1,0 +1,145 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/openclaw/wacli/internal/out"
+	"github.com/openclaw/wacli/internal/store"
+	"github.com/spf13/cobra"
+)
+
+func newCallsCmd(flags *rootFlags) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "calls",
+		Short: "List WhatsApp call events from the local DB",
+	}
+	cmd.AddCommand(newCallsListCmd(flags))
+	return cmd
+}
+
+func newCallsListCmd(flags *rootFlags) *cobra.Command {
+	var chat string
+	var limit int
+	var afterStr string
+	var beforeStr string
+	var asc bool
+
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List call events",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx, cancel := withTimeout(context.Background(), flags)
+			defer cancel()
+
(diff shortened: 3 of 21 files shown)


steipete:
Triage after the 0.9.1 release: PR is still relevant and CI is green, but it needs a quick rebase/fixup before landing.

Evidence:
- Current `main` is already at `0.9.2 - Unreleased` after releasing `0.9.1`.
- A merge simulation of current `main` with this PR is conflict-free, but the PR's changelog entry lands under the already-released `0.9.1 - 2026-05-15` section instead of `0.9.2 - Unreleased`.
- Existing PR checks are from before the release commits, so they do not cover the current base.

Requested fixup:
- rebase or merge current `main`
- move `Calls: persist WhatsApp call signaling and call-log metadata, and add wacli calls list` under `0.9.2 - Unreleased`
- rerun the full gate / let CI refresh

After that this looks like the next PR to review/land for #231.

PR #246, opened 2026-05-16
fix: unwrap edited messages during history sync parsing

### LINKED ISSUE #245: Edited messages from history/backfill are stored as "(message)" instead of unwrapping MESSAGE_EDIT
## Summary

Edited WhatsApp messages that arrive through history sync / on-demand backfill are stored in SQLite as empty text with `display_text = "(message)"`, even though the WhatsApp UI shows the edited message body.

This appears to affect history/backfill parsing specifically. Live message parsing seems to unwrap edits via whatsmeow, but `wacli`'s history parser does not appear to handle `ProtocolMessage_MESSAGE_EDIT`.

## Observed behavior

After syncing/backfilling a group chat containing edited messages:

```sql
SELECT msg_id, datetime(ts, 'unixepoch') AS utc, sender_name,
       length(text), text, display_text
FROM messages
WHERE chat_jid = '<redacted_group_jid>'
  AND ts BETWEEN strftime('%s','2026-05-15T19:20:00Z')
             AND strftime('%s','2026-05-15T19:45:00Z');
```

Representative result:

```text
<msg_id_1>|2026-05-15 19:30:25|<sender>|NULL||(message)
<msg_id_2>|2026-05-15 19:37:01|<sender>|NULL||(message)
```

In the WhatsApp UI, both rows are visible as edited text messages with normal message bodies.

## Expected behavior

`wacli messages list --json`, SQLite storage, and search/export should expose the edited message body, not `(message)`, so edited messages are readable and searchable.

## Suspected root cause

The live path goes through whatsmeow's `ParseWebMessage`, which unwraps edited messages:

```go
if evt.Message.GetProtocolMessage().GetType() == waE2E.ProtocolMessage_MESSAGE_EDIT {
    evt.Info.ID = evt.Message.GetProtocolMessage().GetKey().GetID()
    evt.Message = evt.Message.GetProtocolMessage().GetEditedMessage()
}
```

But the history/backfill path in `internal/app/sync_events.go` calls:

```go
pm := wa.ParseHistoryMessage(chatID, m.Message)
```

That parser eventually reaches `internal/wa/messages.go`. `extractProtocolMutation()` handles `ProtocolMessage_REVOKE`, but does not appear to handle `ProtocolMessage_MESSAGE_EDIT`.

## Proposed fix

For history/backfill messages:

1. Detect `ProtocolMessage_MESSAGE_EDIT`.
2. Use `ProtocolMessage.Key.ID` as the target/original message ID.
3. Parse `ProtocolMessage.EditedMessage` as the actual message content.
4. Store/update the message text/display text so `messages list/search/export` show the edited body.
5. Add regression coverage for a history-sync `WebMessageInfo` containing a `ProtocolMessage_MESSAGE_EDIT`.

## Related prior work

Possibly related but not merged:

- #64
- #80

Related class of `(message)` issues that were accepted/fixed:

- #214
- #231

hiasinho:
## Summary

- Handle edited WhatsApp messages in the history/backfill parser
- Unwrap both `ProtocolMessage_MESSAGE_EDIT` and top-level `EditedMessage`
- Preserve the original target message ID from the protocol key where available
- Add regression coverage for edited messages received through history sync

Fixes #245.

## Behavior note

This fixes parsing for edited messages when they are received by `wacli` during future history sync, fresh initial sync, or backfill events.

It does not automatically repair existing rows already stored as:

```text
text = NULL
display_text = "(message)"
```

Those rows need to be re-delivered by WhatsApp, for example through a fresh linked-device sync, before this parser fix can recover the edited body.

## Testing

```sh
go test ./internal/wa ./internal/app ./internal/store ./cmd/wacli
```

Also tested against a fresh linked-device sync: edited messages that were previously stored as `(message)` were recovered with their text bodies.


hiasinho:
One possible follow-up to make this PR stronger: add regression assertions for the two edge cases handled here too — preserving wrapper `MessageContextInfo` when unwrapping top-level `EditedMessage`, and not overwriting `FromMe` when the protocol key omits it. I can add those if helpful.

PR #247, opened 2026-05-16
feat: add status message sending

dovocoder:
## Summary
- Add `wacli send status` for WhatsApp status broadcasts via `status@broadcast`.
- Support text statuses with optional background color and font metadata.
- Support media statuses with `--file`, optional `--mime`, and `--message` as the media caption.
- Persist synced and locally sent statuses in a dedicated `status_messages` table, including sender fields, media metadata, and text-status styling fields.
- Keep statuses out of normal chat rows and `messages` list/search/export behavior.
- Update the changelog, README, send docs, quickstart, overview, spec, store/sync notes, and integration SQL examples for the new status behavior.

## Test Plan
- `go test ./cmd/wacli ./internal/store ./internal/app`
- `node scripts/build-docs-site.mjs`
- `git diff --check upstream/main..HEAD`

## Notes
- Status sends use WhatsApp status privacy. Targeted recipient selection is not exposed as a CLI flag in this PR.
- Full `pnpm test` was run earlier and still fails on `internal/fsutil TestEnsureWritableDirRejectsNonWritableDir` when executed as root, because root can write to the chmod-restricted test directory. The status-related cmd/store/app tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-- Send: add `wacli send status` for WhatsApp status broadcasts, including text statuses with optional background/font and media statuses with captions.
+- Send: add `wacli send status` for WhatsApp status broadcasts, including text statuses with optional background/font and media statuses with captions. (#247 - thanks @dovocoder)
 - Store: persist synced and locally sent status broadcasts separately in `status_messages` instead of mixing them into normal chat messages.
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## 0.9.2 - Unreleased
 
+### Added
+
+- Send: add `wacli send status` for WhatsApp status broadcasts, including text statuses with optional background/font and media statuses with captions.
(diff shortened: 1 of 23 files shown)


dinakars777 (commented):
Nice work — code quality looks solid. Single clean commit, proper regression tests for both text and media statuses, idempotent schema migration, and clean separation from the regular `messages` table. CI is all green.

A couple of observations before we merge:

1. **Migration placement:** The new columns are added via `migrateStatusMessageMediaColumns` inside `migrateCoreSchema` rather than as a numbered migration in the existing `migrations.go` sequence. Worth checking if Peter prefers the numbered approach for new tables.

2. **Scope check:** This is a sizeable feature (+609 lines, new DB table, new CLI command). @steipete — does status message support fit the current roadmap? Happy to merge once you give the thumbs up.

dovocoder:
Hi @dinakars777, totally agree. Let's wait for confirmation. The reason I added it is that I saw status updates in my messages list yesterday and had to check who that contact was because I hadn't had a conversation with that person in a long time.

PR #248, opened 2026-05-16
fix: align poll compatibility handling

dovocoder:
## Summary
- make outbound poll encoding follow the newer WhatsApp poll message variants for better future compatibility with current WhatsApp clients
- send single-select polls with the V3 poll creation field while keeping multi-select polls on the base field
- use V2 for community announcement groups when live group metadata identifies the target as both announce-only and a community parent
- preserve unmatched poll vote hashes and expose them in `poll show --json` so newer/unknown poll options do not disappear silently
- update poll send docs and changelog

## Why
WhatsApp has multiple poll creation fields in the wire format. The existing whatsmeow convenience helper only emits the original/base field, while other clients already select newer fields depending on poll type. This PR keeps wacli aligned with those message variants so outbound polls remain compatible as WhatsApp clients continue to evolve.

This also makes poll result inspection more forward-compatible: if a synced vote references an option hash that wacli cannot currently map to a stored option, the hash is preserved and surfaced in JSON instead of being dropped.

## Notes
- whatsmeow's high-level `BuildPollCreation` helper only builds the base poll field, but the generated proto exposes V2/V3 fields, so wacli now builds the minimal poll creation message locally and still sends through `SendMessage`.
- No broad poll metadata migration was added; unknown hashes are stored in the existing poll vote JSON payload only when needed.
- V2 selection is intentionally conservative and depends on available live group metadata, avoiding speculative behavior for regular groups.

## Test Plan
- [x] `git diff --check upstream/main...HEAD`
- [x] `go test ./...`
- [x] `pnpm test`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixed
 
 - History: unwrap edited WhatsApp messages during history sync and backfill so stored/searchable text shows the edited body instead of `(message)`. (#246 - thanks @hiasinho)
+- Polls: use WhatsApp's single-select poll creation field for outbound single-select polls and preserve unmatched poll vote hashes in `poll show --json`.
 - Sync: canonicalize `@lid` chat JIDs before enqueuing media downloads so `sync --follow --download-media` finds the correct DB row for live one-to-one messages. (#244 - thanks @Daniel1of1)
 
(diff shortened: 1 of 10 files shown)


dinakars777:
Thanks @dovocoder

PR #249, opened 2026-05-17
ci: add aqua-friendly release assets

dovocoder:
## Summary
- Publish aqua-friendly release archives for Darwin, Linux, and Windows while keeping the Darwin universal archive.
- Replace split platform checksum assets with one combined `checksums.txt` covering every release archive.
- Update the Homebrew tap handoff to use target-specific release binaries for macOS and Linux.

## Release assets
- `wacli-darwin-amd64.tar.gz`
- `wacli-darwin-arm64.tar.gz`
- `wacli-darwin-universal.tar.gz`
- `wacli-linux-amd64.tar.gz`
- `wacli-linux-arm64.tar.gz`
- `wacli-windows-amd64.zip`
- `checksums.txt`

## Validation
- `go run github.com/goreleaser/goreleaser/v2@v2.15.4 check --config .goreleaser.yaml`
- `go run github.com/goreleaser/goreleaser/v2@v2.15.4 check --config .goreleaser-linux-windows.yaml`
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/release.yml`
- Python YAML parse for release and GoReleaser YAML files
- Shell checksum-generation smoke test for the six expected archive names
- `git diff --check`
- Dry-run release workflow on fork: https://github.com/dovocoder/wacli/actions/runs/25992692978
  - verified all expected assets were published
  - verified `checksums.txt` contains six archive entries
  - cleaned up the temporary test release/tag afterward

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,7 +53,7 @@ jobs:
 
   update-homebrew-tap:
     runs-on: ubuntu-latest
-    needs: goreleaser-darwin
+    needs: goreleaser-linux-windows
     steps:
       - name: Resolve release tag
         env:
@@ -85,7 +85,9 @@ jobs:
             -f formula=wacli \
             -f tag="$RELEASE_TAG" \
             -f repository=openclaw/wacli \
-            -f macos_artifact=wacli-macos-universal.tar.gz \
+            -f description="WhatsApp CLI built on whatsmeow" \
+            -f artifact_template=wacli-{target}.tar.gz \
+            -f target_aliases=darwin_amd64=darwin-amd64,darwin_arm64=darwin-arm64,linux_amd64=linux-amd64,linux_arm64=linux-arm64 \
             -f request_id="$request_id"
 
           run_id=""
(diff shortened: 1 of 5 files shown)


dinakars777:
Thanks, the release-asset side looks useful and the GoReleaser/actionlint checks pass. I don’t think this is ready yet because the Homebrew tap handoff does not appear to be validated against the current tap updater/formula shape.

The current `Formula/wacli.rb` still has one macOS universal URL and one Linux source URL shape. The tap updater only uses `artifact_template` for formulas that already have per-target URLs, so this dispatch is likely to fall back to the old macOS/source update path. Since this PR also renames the universal asset from `wacli-macos-universal.tar.gz` to `wacli-darwin-universal.tar.gz`, that fallback path can look for a release asset that no longer exists.

The fork dry-run also skipped the real tap update because `HOMEBREW_TAP_TOKEN` was not configured.

Can you either update/validate the tap side so `artifact_template` converts the current wacli formula to target-specific assets, or keep the release workflow compatible with the current tap updater? A good validation would show the exact resulting `Formula/wacli.rb` diff from the tap workflow.

dovocoder:
Hi sure, I will check it

dovocoder:
Depended on https://github.com/steipete/homebrew-tap/pull/39

steipete:
Thanks for pushing this forward. I ended up landing the release/tap work directly on main with the repo-wide convention we already use for Go tools:

- wacli release assets now use target-specific names like `wacli_<version>_darwin_arm64.tar.gz`, `wacli_<version>_linux_amd64.tar.gz`, and one combined `checksums.txt`.
- The release workflow now dispatches `openclaw/homebrew-tap` instead of the old personal tap.
- The OpenClaw tap now has a stronger updater that can convert the old wacli formula shape from shared macOS/source URLs to target-specific binary URLs.

Landed in:

- openclaw/wacli@19d0ef4
- openclaw/homebrew-tap@b51b534
- steipete/homebrew-tap@d62b493 and steipete/homebrew-tap@c74a103

Closing this PR as superseded by those commits. Thanks again for identifying the Aqua/release-asset gap.

PR #250, opened 2026-05-18
feat: allow read-only media downloads

mbelinky:
## Summary

- allow `wacli media download` to run in `--read-only` mode when an explicit `--output` path is provided
- download and decrypt media from synced metadata without opening the WhatsApp session store or taking the write lock
- leave the existing writable download path unchanged, including DB `local_path` / `downloaded_at` recording

## Why

Long-running `wacli sync --follow` processes can legitimately hold the store lock. In that state, already-synced media metadata is still useful for output-only media extraction, but the previous command path always required writable mode, opened the WhatsApp session store, and failed before it could download.

## Validation

- Manual live check: while a `sync --follow` process held the store lock, normal writable `media download` still failed with the lock, while `media download --read-only --output ...` downloaded audio and video files and left stored media metadata unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+- Media: allow `media download --read-only` when `--output` is explicit, so synced media can be fetched without opening the WhatsApp session store or recording download state. (#250 - thanks @mbelinky)
+
 ### Security
 
 ### Fixed
--- a/cmd/wacli/media.go
+++ b/cmd/wacli/media.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/openclaw/wacli/internal/out"
+	"github.com/openclaw/wacli/internal/wa"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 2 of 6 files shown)


dinakars777:
Thanks, this is a useful improvement. I don’t think it is quite ready yet because the new direct media downloader accepts full http:// / https:// URLs from stored media metadata.

whatsmeow’s existing DownloadMediaWithPath path requires a slash-prefixed WhatsApp direct path and constructs the media URL itself. Allowing arbitrary URLs here broadens `media download --read-only` into a generic HTTP fetch path if stored metadata is malformed or tampered with.

Can you keep production behavior strict by rejecting full URLs and only accepting slash-prefixed direct paths? The test can inject a test media host or exercise the lower-level decrypt/download helper without making arbitrary URLs valid production input.

Please also add a short docs/media.md note for the new `--read-only --output` behavior.

steipete:
@dinakars777 that's not really how I'd approach PRs. If you already review it, just fix it up and merge. Saves everyone time. You still act like code is expensive. Once you have it loaded into your agent just fix it up.

I'll do that now.

dinakars777:
> @dinakars777 that's not really how I'd approach PRs. If you already review it, just fix it up and merge. Saves everyone time. You still act like code is expensive. Once you have it loaded into your agent just fix it up.
> 
> I'll do that now.

@steipete - I will follow this model going forward.

PR #261, opened 2026-05-22
fix: import completed call logs

### LINKED ISSUE #256: call_events only captures 'offer' — no duration/outcome from sync --once
## Summary

`wacli calls list` only ever contains `offer` events. No `accept`, `terminate`, `reject`, or historical `CallLogMessage` data is persisted. Duration and outcome are always null.

This appears to be a `sync --once` limitation: the `offer` event is queued on the server before the call starts and gets picked up during the next sync, but `accept`/`terminate` events happen _during_ the active call when wacli isn't connected.

## Evidence

Three inbound calls this month from the same contact, including one that lasted 1h 18m (verified via WhatsApp UI). All three are offer-only:

```json
[
  {"event_type": "offer", "direction": "inbound", "media": "audio", "outcome": null, "reason": null, "duration_secs": null, "timestamp": "2026-05-20T20:23:20Z"},
  {"event_type": "offer", "direction": "inbound", "media": "audio", "outcome": null, "reason": null, "duration_secs": null, "timestamp": "2026-05-19T19:53:42Z"},
  {"event_type": "offer", "direction": "inbound", "media": "audio", "outcome": null, "reason": null, "duration_secs": null, "timestamp": "2026-05-19T18:55:09Z"}
]
```

## Expected

At minimum, duration and outcome should be populated for completed calls. The PR #242 implementation handles both live events and historical `CallLogMessage` payloads — the question is whether `sync --once` triggers the app-state sync path that surfaces call-log history.

## Possible paths

1. **App-state call logs:** If `CallLogMessage` entries exist in app-state but aren't fetched during `sync --once`, a dedicated `calls sync` or `--include-call-history` flag on sync could populate them retroactively.
2. **Post-call backfill:** A `sync --once` run _after_ a completed call could check for `terminate`/historical events that arrived since last sync.
3. **Documentation:** If this is an inherent limitation of the non-persistent connection model, document it so downstream tools can infer duration from timestamps (e.g., offer → next message gap).

## Context

Using `sync --once --idle-exit 10s` in a cron-based event collector (runs every ~2 min). The downstream consumer (ekku event pipeline) needs call duration to distinguish "had a 1h18m phone call" from "phone rang once" — currently both look identical.

wacli v0.9.2

dovocoder:
## Summary

Fixes #256 by importing completed WhatsApp call logs from the app-state/history paths that actually carry them, instead of relying only on live call signaling events.

## What changed

- Replay the `regular` app-state collection during sync so existing call-log records are emitted and stored idempotently.
- Store `HistorySync.callLogRecords` when WhatsApp includes call logs in history sync payloads.
- Decode app-state call records correctly:
  - `startTime` can be seconds, milliseconds, or microseconds
  - self identity prefers the linked-device LID for app-state call records
  - one-to-one records resolve to the peer chat, not the account’s own JID
- Make call-event upserts correct previously imported rows for the same call ID, including timestamp/chat fixes.

## Root cause

`calls list` mostly showed live `offer` events because completed call logs arrive as call-log records in WhatsApp app-state/history data. The `regular` app-state collection was not being replayed, so if WhatsApp had already advanced that collection before wacli knew how to store call logs, later delta syncs could not recover those records. The call-log parser also assumed millisecond timestamps and PN self identity, which made replayed records land under 1970/self-chat in local testing.

## Local verification

Full local gate passed:

```text
```

Redacted live proof from a local store:

Before the fix, after sync:

```json
{"success":true,"data":{"calls":null},"error":null}
```

After the fix, `sync --once` imported the same-day completed call as a structured call-log row:

```json
{
  "chat_jid": "<peer>@s.whatsapp.net",
  "chat_name": "<redacted>",
  "sender_jid": "<self>@s.whatsapp.net",
  "sender_name": "<redacted>",
  "call_id": "0002...1BC",
  "event_type": "call_log",
  "direction": "outbound",
  "media": "audio",
  "outcome": "connected",
  "call_type": "regular",
  "duration_secs": 9,
  "timestamp": "2026-05-22T07:44:30Z",
  "participants": [
    {"jid": "<peer>@s.whatsapp.net", "outcome": "connected"}
  ]
}
```

This proves the missing completed-call case from #256: the imported row is no longer just an `offer`; it has the peer chat, real timestamp, duration, direction, media, and connected outcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,11 @@
 ### Added
 
 - Chats: store unread marker state and numeric `unread_count` separately; migrate existing stores away from sentinel unread values while preserving public chat JSON fields. (#255 - thanks @drelum and @dovocoder)
-
 ### Security
 
 ### Fixed
+
+- Calls: import call-log records from full history syncs and the `regular` app-state collection so old call events can be stored even when the live call signaling was missed. (#256)
 
 ### Docs
 
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -83,6 +83,7 @@ type WAClient interface {
 	RequestAppStateRecovery(ctx context.Context, name string) (types.MessageID, error)
 	Logout(ctx context.Context) error
 	LinkedJID() string
+	LinkedLID() string
 
 	SetProfilePicture(ctx context.Context, avatar []byte) (string, error)
 }
--- a/internal/app/fake_wa_test.go
+++ b/internal/app/fake_wa_test.go
@@ -26,6 +26,7 @@ type fakeWA struct {
 
 	authed    bool
 	connected bool
+	linkedLID string
 
 	nextHandlerID uint32
 	handlers      map[uint32]func(interface{})
@@ -652,4 +653,13 @@ func (f *fakeWA) LinkedJID() string {
 		return ""
 	}
 	return "1234567890@s.whatsapp.net"
+}
+
+func (f *fakeWA) LinkedLID() string {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	if !f.authed {
+		return ""
+	}
+	return f.linkedLID
 }
--- a/internal/app/sync.go
+++ b/internal/app/sync.go
@@ -178,8 +178,9 @@ func (a *App) Sync(ctx context.Context, opts SyncOptions) (SyncResult, error) {
 }
 
 func (a *App) syncAppStateDeltas(ctx context.Context) {
-	for _, name := range []appstate.WAPatchName{appstate.WAPatchRegularHigh, appstate.WAPatchRegularLow} {
-		if err := a.wa.FetchAppState(ctx, string(name), false, false); err != nil {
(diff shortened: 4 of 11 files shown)


dovocoder:
@clawsweeper re-review

dovocoder:
Addressed the ClawSweeper finding.

What changed in the updated single commit:

- live call signaling events now use the PN `LinkedJID` identity for `ParseLiveCallEvent` direction checks
- app-state/history call-log records still use the LID-first identity path for `CallLogRecord` parsing
- added regression coverage for an outbound live call offer when `LinkedLID` is present, so it remains `direction: outbound`

Acceptance checks passed:

```text
go test ./internal/app ./internal/wa ./internal/store
pnpm format:check
git diff --check
```

@clawsweeper re-review

steipete:
Pushed a maintainer fixup in a407a2b.

What changed:
- scoped call-event correction updates to one existing row in the same chat, avoiding cross-chat movement for duplicate call IDs
- passed both linked LID and linked PN identities into app-state/history call-log parsing
- added regressions for duplicate call IDs across chats and PN/LID self matching

Local proof:
- `go test ./internal/app ./internal/wa ./internal/store`
- autoreview local patch clean: no accepted/actionable findings

@clawsweeper re-review

PR #272, opened 2026-06-04
feat(presence): delegate typing/paused through the send socket when the store is locked

### LINKED ISSUE #239: Feature request: delegate presence typing/paused through sync --follow socket
## Problem

When `wacli sync --follow` is running, it holds the store lock continuously. The `send` commands (text, file, react, etc.) correctly delegate through the Unix socket at `<store>/.send.sock`, but `presence typing` and `presence paused` do **not** — they attempt to acquire the store lock directly and fail:

```
store is locked (another wacli is running?): store locked: resource temporarily unavailable
```

This makes it impossible to send typing indicators while continuously syncing, which is a common use case for bots and relay scripts.

## Requested Change

Add `presence typing` and `presence paused` to the IPC delegation mechanism in `send_ipc.go` — similar to how `send react` is already delegated.

Suggested addition to `sendDelegateRequest`:
- `Kind: "presence_typing"` / `Kind: "presence_paused"`
- Fields: `To string`, `Media string` (for audio indicator)

## Workaround

Currently using a `send react --reaction 👀` as a poor-man's "received" indicator, cleared with an empty reaction when the reply is sent. Works but is ugly.

## Context

Using wacli as a relay bridge (wacli-relay) that forwards incoming WhatsApp messages to a webhook and sends replies back. The typing indicator is important UX feedback while the webhook processes the message. Signal-cli (the Signal equivalent) supports this natively via DBus.

kidshaker:
## Problem

`wacli presence typing|paused` fails with a store-lock error whenever a `sync --follow` daemon is running. The presence command opens its own client and tries to take the exclusive store lock, but the daemon holds that lock continuously, so presence always errors with `store is locked`.

The `send` family of commands already solves this: when the lock is held they delegate the operation to the running daemon through the `.send.sock` Unix socket. Chat presence was never wired into that path, so the WhatsApp "typing…" indicator can't be sent while the sync daemon runs.

## Change

Extends the existing send-delegation mechanism to chat presence, mirroring the `react` path exactly:

- `send_ipc.go`: adds `presence_state` / `presence_media` fields to `sendDelegateRequest`, a `case "presence"` in `executeDelegatedSend`, and an `executeDelegatedPresence` handler that calls `a.WA().SendChatPresence(...)` on the daemon side.
- `presence.go`: when `newApp` fails on a held lock, `runPresence` now calls `tryDelegateSend` (same pattern as `send react`) instead of returning the lock error.
- `send_ipc_test.go`: JSON round-trip test for the new presence fields.

No behavior change when no daemon socket is present: `tryDelegateSend` only triggers when the error is a lock error **and** the socket exists, otherwise the original error is returned unchanged.

## Testing

- `go build ./...` ✓
- `go vet ./cmd/wacli/` ✓
- `go test ./cmd/wacli/` ✓ (existing suite + new test)

Closes #239

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Groups: add live admin commands for creating groups, setting descriptions, toggling announce-only and admin-only edits, and approving or rejecting join requests. (#265 - thanks @dovocoder)
 - Groups: include resolved phone numbers in pending join request output when LID mappings are available. (#273 - thanks @cielecki)
+- Presence: delegate typing and paused indicators through the sync daemon send socket when the store lock is held. (#272 - thanks @kidshaker)
 - Profile: add commands to remove the profile picture, set About text, set the profile display name, fetch profile picture metadata, fetch a user's About text, and fetch WhatsApp Business profile details. (#267 - thanks @dovocoder)
 - Sync: add opt-in keepalive-failure stale detection for `sync --follow` (`1s` to `<2m20s`), including forced reconnect, a `stale` NDJSON event, a private store `HEARTBEAT`, and `doctor --json` `last_activity_at`. (#278 - thanks @thedavidweng)
 
--- a/cmd/wacli/presence.go
+++ b/cmd/wacli/presence.go
@@ -137,3 +137,14 @@ func presenceMediaFromString(media string) (types.ChatPresenceMedia, error) {
 		return "", fmt.Errorf("unsupported --media %q (supported: audio)", media)
 	}
 }
+
+func presenceStateFromString(state string) (types.ChatPresence, error) {
+	switch strings.ToLower(strings.TrimSpace(state)) {
+	case string(types.ChatPresenceComposing):
+		return types.ChatPresenceComposing, nil
+	case string(types.ChatPresencePaused):
+		return types.ChatPresencePaused, nil
+	default:
+		return "", fmt.Errorf("unsupported presence state %q", state)
+	}
+}
--- a/cmd/wacli/presence.go
+++ b/cmd/wacli/presence.go
@@ -67,6 +67,18 @@ func runPresence(flags *rootFlags, to string, state types.ChatPresence, media st
 
 	a, lk, err := newApp(ctx, flags, true, false)
 	if err != nil {
+		resp, delegated, delegateErr := tryDelegateSend(ctx, flags, err, sendDelegateRequest{
+			Kind:          "presence",
+			To:            to,
+			PresenceState: string(state),
+			PresenceMedia: media,
+		})
+		if delegated {
+			if delegateErr != nil {
+				return delegateErr
+			}
+			return writePresenceOutput(flags, state, resp)
+		}
 		return err
 	}
 	defer closeApp(a, lk)
@@ -100,6 +112,18 @@ func runPresence(flags *rootFlags, to string, state types.ChatPresence, media st
 		})
 	}
 	fmt.Fprintf(os.Stdout, "Presence '%s' sent to %s\n", state, toJID.String())
+	return nil
+}
+
+func writePresenceOutput(flags *rootFlags, state types.ChatPresence, resp sendDelegateResponse) error {
+	if flags.asJSON {
+		return out.WriteJSON(os.Stdout, map[string]any{
+			"sent":  true,
+			"to":    resp.To,
+			"state": string(state),
+		})
+	}
+	fmt.Fprintf(os.Stdout, "Presence '%s' sent to %s\n", state, resp.To)
 	return nil
 }
 
--- a/cmd/wacli/presence.go
+++ b/cmd/wacli/presence.go
@@ -100,7 +100,9 @@ func runPresence(flags *rootFlags, to string, state types.ChatPresence, media st
 		return err
 	}
 
-	if err := a.WA().SendChatPresence(ctx, toJID, state, chatMedia); err != nil {
+	if err := sendPresenceWithRetry(ctx, reconnectForSend(a), func(ctx context.Context) error {
+		return a.WA().SendChatPresence(ctx, toJID, state, chatMedia)
+	}); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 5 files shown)


steipete:
Maintainer update on head 238151ff66fcf22b0025ffd0664d9477745e62c6: I strengthened the branch, but I am intentionally not merging it yet.

What changed beyond the original patch:
- Added a real locked-store/send-socket e2e regression using a subprocess CLI client, an actual held wacli store lock, and a Unix .send.sock test daemon. It proves both `presence typing --media audio` and `presence paused` delegate through the socket while the direct store is locked.
- The e2e asserts the delegated request kind/state/media/recipient, propagates `--timeout` as `timeout_ms`, returns delegated JSON output, and does not fall through to unauthenticated/not-connected direct client behavior.
- Added locked-store fallback proof with no socket: the command returns the original lock error and does not leak the delegate-unavailable error.
- Added daemon-error propagation proof: a socket response with `ok:false` returns the daemon error instead of the original lock error.
- Hardened daemon-side presence IPC to reject unsupported presence states, and made direct/delegated presence use the established send retry/reconnect path for transient `not connected` windows.
- Added the maintainer changelog entry with @kidshaker credit.

Local verification:
- go test ./cmd/wacli -run 'TestPresence|TestExecuteDelegatedSendRoutesPresence|TestSendPresenceWithRetry|TestTryDelegateSend|TestSendDelegateRequestPreservesPresence'
- pnpm -s docs:site && pnpm -s format:check && pnpm -s lint && pnpm -s test && pnpm -s build && git diff --check
- go test -race ./cmd/wacli ./internal/app ./internal/wa ./internal/fsutil
- go run golang.org/x/tools/cmd/deadcode@v0.45.0 -test -tags sqlite_fts5 ./...
- goreleaser check --config .goreleaser.yaml && goreleaser check --config .goreleaser-linux-windows.yaml
- autoreview --mode branch --base origin/main: clean, no accepted/actionable findings

CI on 238151ff66fcf22b0025ffd0664d9477745e62c6 passed: test, linux-release-builds, docker.

Live-provider blocker: a safe authenticated local WhatsApp account is not available in this checkout. Redacted read-only checks show accounts=[] and the default store authenticated=false, connected=false. Because issue #239 was withdrawn and the prior review's material requirement is real locked-daemon WhatsApp presence behavior, the remaining owner decision is: provide a safe live account/message target for presence proof, or close #272 as withdrawn/unproven. I am not closing or merging it autonomously.

steipete:
Live proof attempt on current head `238151ff66fcf22b0025ffd0664d9477745e62c6`:

- Rebuilt and used the real candidate binary from this exact PR head.
- Confirmed the local `me` account is authenticated; linked and target JIDs redacted.
- Cleared a stale auth-process store lock, then started the candidate daemon. The daemon connected, held the store lock, and owned `.send.sock` with socket permissions `srw-------`.
- From a separate process while the store was locked by the daemon, `presence typing` returned success for the redacted target through the send socket.
- Then `presence paused` returned success for the same redacted target through the send socket.
- Clean shutdown verified: daemon stopped, store lock released, `.send.sock` removed, and no daemon process remained.

Independent UI observation limitation:

- I used the safest available recipient first: the authenticated account's own WhatsApp self-chat in the real WhatsApp desktop client.
- Self-chat opened as `Message yourself`; both the typing and paused screenshots showed no visible typing indicator in that chat.
- That means this is real locked-daemon/socket proof, but not meaningful recipient-side presence observation.

I am not merging on this proof. The exact remaining requirement is a safe second WhatsApp account/device/chat where the recipient-side real WhatsApp client can observe the sender's typing indicator after `typing` and its disappearance after `paused`. Once that recipient is available, this should be rerun against this PR head before merge.

steipete:
Owner waiver and final proof for head `238151ff66fcf22b0025ffd0664d9477745e62c6`:

- Owner decision: no safe second WhatsApp recipient/account/device exists for this item, so independent recipient-side typing-indicator observation is explicitly waived for #272 only.
- This waiver does not lower the normal proof bar for future provider-facing presence changes; it accepts the strongest safe evidence available for this specific PR.

Accepted evidence before merge:

- Exact final candidate commit: `238151ff66fcf22b0025ffd0664d9477745e62c6`.
- Real authenticated local WhatsApp account used; linked and target JIDs redacted.
- Real candidate binary rebuilt from that head and used for live proof.
- Candidate daemon connected, held the real store lock, and owned the real `.send.sock` with socket permissions `srw-------`.
- Separate client process sent `presence typing` while the store was locked by the daemon; the command succeeded through `.send.sock`.
- Separate client process then sent `presence paused` while the store was locked by the daemon; the command succeeded through `.send.sock`.
- Safest available UI attempt was WhatsApp self-chat in the real WhatsApp desktop client; it showed `Message yourself`, so it did not provide meaningful recipient-side typing visibility.
- Clean shutdown verified: daemon stopped, store lock released, `.send.sock` removed, and no daemon process remained.
- Non-provider e2e proof covers held store lock + Unix `.send.sock` dispatch for `typing --media audio` and `paused`, no direct second connection path, daemon error propagation, timeout propagation, JSON output, and no-socket fallback semantics.
- Local gate previously passed on this exact head: focused presence/socket tests, full `pnpm -s docs:site && pnpm -s format:check && pnpm -s lint && pnpm -s test && pnpm -s build && git diff --check`, race checks, deadcode, GoReleaser config checks, and autoreview with no accepted/actionable findings.
- Current-head CI is green: `test`, `linux-release-builds`, and `docker`.

Merging with contributor credit preserved from @kidshaker's authored feature commit.

steipete (approved):
Approved with owner waiver recorded and current-head CI green.

steipete:
Landed with the item-specific owner waiver recorded above.

- PR head tested: `238151ff66fcf22b0025ffd0664d9477745e62c6`
- Landed commits on `main`: `fdc7120882da8a9f2d13df0586b410f54dc0abac`, `3731f0f`, `1cdf61d514d14866a8825cd27cf24d77a436801d`
- Contributor credit preserved: the feature commit remains authored by @kidshaker; the changelog thanks @kidshaker.
- Live proof: authenticated real account, exact candidate binary, real daemon holding the store lock, owner-only `.send.sock`, successful delegated `typing` then `paused`, self-chat observation attempt, and clean daemon/socket/lock shutdown.
- Local proof: focused socket/presence tests; full docs/format/vet/plain+FTS tests/build gate; race subset; deadcode; both GoReleaser config checks; autoreview clean.
- Current-head PR CI passed: `test`, `linux-release-builds`, and `docker`.
- Released in [v0.11.1](https://github.com/openclaw/wacli/releases/tag/v0.11.1).

Thank you, @kidshaker.

PR #282, opened 2026-06-18
fix(sync): send available presence on connect

dovocoder:
## Problem

WhatsApp requires an `<presence type="available"/>` stanza after a link to mark the device as active. Without it the connection stays alive, but the mobile app shows a stale "last active" timestamp.

The latest upstream `main` added `sendInitialAvailablePresence` in `internal/wa/client.go`, but it only runs once inside the initial `ConnectContext` call and only when the pushname is already set. It therefore misses two real-world cases:

- **Reconnects inside a running sync session** fire `*events.Connected` again, but `ConnectContext` is not re-entered, so no presence is re-sent.
- **Late-arriving pushname** is delivered as `*events.PushNameSetting` after connect. If the pushname was empty at connect time, the upstream path never sends presence.

## Solution

Keep the upstream early-connect attempt and add an event-loop path that covers the cases above.

### Files changed

- `internal/wa/client.go` — adds `(Client).SendPresence()`, a concurrency-safe wrapper around `whatsmeow.Client.SendPresence()`.
- `internal/app/app.go` — adds `SendPresence()` to the `WAClient` interface.
- `internal/app/sync_events.go` — on `*events.Connected` and `*events.PushNameSetting`, calls `go a.sendPresence(...)`.
- `internal/app/sync_events.go` — adds `App.sendPresence()` helper that logs failures as NDJSON warning events (`send_presence_failed`) and never aborts sync.
- `internal/app/fake_wa_test.go` — adds `SendPresence()` fake implementation and tracking fields so the new behavior is unit-testable.
- `internal/app/sync_presence_test.go` — adds regression tests for connected, push-name setting, and presence-error paths.

## Tests

The full gate was run locally and passes:

```bash
```

New regression tests:

```bash
go test ./internal/app -run TestSyncSendsAvailablePresence -count=1
go test ./internal/app -run TestSyncPresenceFailureWarnsAndContinues -count=1
```

## Real behavior proof

A live linked-device run confirms the new event-loop path sends `PresenceAvailable` through the wrapper. Redacted stderr/NDJSON log is attached in [this comment](https://github.com/openclaw/wacli/pull/282#issuecomment-4742536673).

## Deployment / rollout

No operational changes required. The compiled `dist/wacli` binary is already available at `~/.local/bin/wacli`.

## Checklist

- [x] Code compiles / builds.
- [x] Tests pass (`pnpm test` and FTS path).
- [x] Lint / formatting checks pass.
- [x] `git diff --check` clean.
- [x] Branch rebased onto latest `openclaw/wacli:main` (`464254a`).
- [x] Redacted live WhatsApp behavior proof added to PR comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Fixed
 
 - WhatsApp connectivity: update `whatsmeow` for the current WhatsApp protocol and fix `405 (Client Outdated)` failures. (#280)
-- Sync: send available presence on reconnect and when the server sends the pushname after the initial connect, so the linked device shows as active even when the pushname arrives late. (#282 - thanks @dovocoder)
+- Sync: send available presence on reconnect and when the server sends the pushname after the initial connect, and send unavailable presence after a successful sync so the phone continues to receive push notifications. (#282 - thanks @dovocoder)
 
 ## 0.11.1 - 2026-06-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - WhatsApp connectivity: update `whatsmeow` for the current WhatsApp protocol and fix `405 (Client Outdated)` failures. (#280)
+- Sync: send available presence on reconnect and when the server sends the pushname after the initial connect, so the linked device shows as active even when the pushname arrives late. (#282 - thanks @dovocoder)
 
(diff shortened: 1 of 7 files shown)


dovocoder:
Superseded by #283, which combines these changes with the websocket context fix into a single PR. All commits from this PR are included in #283's squashed commit.

PR #283, opened 2026-06-19
fix(sync): presence handling for linked-device active status and phone push notifications

dovocoder:
## Problem

A `wacli sync` linked device stays connected but never tells WhatsApp it is active, and when sync exits it fails to tell WhatsApp it is inactive. This causes two user-visible issues:

1. **Stale "last active" timestamp** — WhatsApp's Linked Devices screen shows the device as inactive even while `sync --follow` is running.
2. **Phone stops receiving push notifications** — after sync exits, the device stays marked "available" because the `<presence type="unavailable"/>` send fails every time.

### Root causes

**Missing available presence.** The upstream `main` added `sendInitialAvailablePresence` in `ConnectContext`, but it only runs once at initial connect and only when the pushname is already set. It misses reconnects and late-arriving pushname events.

**Unavailable presence fails on SIGTERM.** `wa.Client.Connect` passed the caller's (signal-cancellable) context to `cli.ConnectContext(ctx)`. When the signal handler cancels it (e.g., `timeout` sending SIGTERM), whatsmeow closes the websocket before `sync.go:207` can send the unavailable presence.

### 1. Send available presence on connect and pushname

On `*events.Connected` and `*events.PushNameSetting`, call `sendPresence(ctx, types.PresenceAvailable)` synchronously (not in goroutines) so a delayed available send cannot run after the final unavailable cleanup send.

### 2. Send unavailable presence after sync exit

After a successful sync run, call `sendPresence(ctx, types.PresenceUnavailable)` so the phone resumes push notifications. The event handler is removed before this send so no late `Connected`/`PushNameSetting` event can fire an available presence after cleanup. The cleanup send uses a 5-second timeout context so a stalled WhatsApp write cannot block shutdown.

### 3. Detach websocket context for sync only with bounded dial

Add a `DetachSocket bool` field to `ConnectOptions`. When `true` (set by `connectForSync` and `ReconnectWithBackoff` — sync only), `ConnectContext` uses a detached context with a **30-second dial timeout** (`time.AfterFunc`). The dial is bounded so a hanging connect cannot outlive SIGTERM indefinitely; after the dial succeeds, the timer is stopped so the websocket stays open until `Disconnect()`/`Close()`. Other commands keep the default `false` so `--timeout` and signal cancellation still bound the connection.

The retry loop in `ReconnectWithBackoff` still uses the passed context for cancellation control — only the websocket lifetime is decoupled.

## Files changed

- `internal/wa/client.go` — adds `DetachSocket` field to `ConnectOptions`; uses a dial-timeout-bounded detached context for `ConnectContext` when `DetachSocket` is true; adds `(Client).SendPresence()` wrapper.
- `internal/app/app.go` — adds `SendPresence()` to the `WAClient` interface.
- `internal/app/sync_events.go` — on `*events.Connected` and `*events.PushNameSetting`, calls `sendPresence(ctx, types.PresenceAvailable)` synchronously; adds `App.sendPresence()` helper.
- `internal/app/sync.go` — after a successful sync run, removes the event handler, then sends unavailable presence with a 5-second timeout context; sets `DetachSocket: true` in `connectForSync`.
- `internal/app/fake_wa_test.go` — adds `SendPresence()` fake implementation and tracking fields.
- `internal/app/sync_presence_test.go` — regression tests for connected, push-name setting, failure-warning, synchronous behavior, and the final unavailable send.

## Tests

The full gate passes locally:

```bash
```

New regression tests:

```bash
go test ./internal/app -run TestSyncSendsAvailablePresence -count=1
go test ./internal/app -run TestSyncPresenceFailureWarnsAndContinues -count=1
go test ./internal/app -run TestSyncSendsAvailablePresenceOnConnectedIsSynchronous -count=1
```

> **Note on websocket-context test coverage:** the unavailable-send-on-SIGTERM fix lives in `wa.Client.Connect` (which calls `cli.ConnectContext`). The app-level tests use a fake WA client whose `SendPresence` always succeeds, so they verify the *call sequence* (available → unavailable) but cannot catch the real `cli.IsConnected()` failure path. A unit test for the detached-context fix would require refactoring `wa.Client` to accept an interface for the whatsmeow client. The fix was verified with a live WhatsApp account (see below).

## Live verification

Tested on a real linked-device account with a `timeout 30s` cron job:

**Before fix — every unavailable send failed:**
```
{"event":"signal","action":"shutdown","signal":"terminated"}
{"event":"stopping","messages_synced":19}
{"event":"warning","code":"send_presence_failed","error":"...context canceled"}
Messages stored: 19
```

**After fix — unavailable send succeeds on SIGTERM:**
```
{"event":"signal","action":"shutdown","signal":"terminated"}
{"event":"stopping","messages_synced":9}
Messages stored: 9
```

No `send_presence_failed` warning — the unavailable presence was sent successfully because the websocket stayed open (DetachSocket with bounded dial).

## Operational notes

No new flags or env vars. The compiled `dist/wacli` binary is already deployed.

For continuous sync via cron, a `timeout 30s wacli sync --idle-exit 10s` cron entry now works correctly: sync runs for ~30s, exits on SIGTERM, sends unavailable presence, and the phone receives push notifications for the remaining ~30s of each minute.

## Checklist

- [x] Code compiles / builds.
- [x] Tests pass (`pnpm test` and FTS path).
- [x] Lint / formatting checks pass.
- [x] `git diff --check` clean.
- [x] Commits squashed into one.
- [x] No release-owned CHANGELOG.md edits.
- [x] Live WhatsApp behavior proof included.
- [x] Available presence sends are synchronous (no goroutine race).
- [x] Event handler removed before unavailable cleanup send.
- [x] Unavailable cleanup send bounded with 5-second timeout.
- [x] `DetachSocket` is sync-only; other commands keep caller-cancellation.
- [x] Detached connect attempt bounded with 30-second dial timeout.

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -81,6 +81,7 @@ type WAClient interface {
 	DownloadMediaToFile(ctx context.Context, directPath string, encFileHash, fileHash, mediaKey []byte, fileLength uint64, mediaType, mmsType string, targetPath string) (int64, error)
 
 	SendChatPresence(ctx context.Context, jid types.JID, state types.ChatPresence, media types.ChatPresenceMedia) error
+	SendPresence(ctx context.Context, presence types.Presence) error
 	ParseWebMessage(chatJID types.JID, webMsg *waWeb.WebMessageInfo) (*events.Message, error)
 	DecryptReaction(ctx context.Context, reaction *events.Message) (*waProto.ReactionMessage, error)
 	SetManualHistorySyncDownload(enabled bool)
--- a/internal/app/fake_wa_test.go
+++ b/internal/app/fake_wa_test.go
@@ -63,10 +63,12 @@ type fakeWA struct {
 	pinCalls                    []fakePinCall
 	muteCalls                   []fakeMuteCall
 	markReadCalls               []fakeMarkReadCall
+	manualHistorySyncCalls      []bool
+	appStateRecoveries          []string
+	appStateFetches             []fakeAppStateFetch
 
-	manualHistorySyncCalls []bool
-	appStateRecoveries     []string
-	appStateFetches        []fakeAppStateFetch
+	presenceCalls   []types.Presence
+	sendPresenceErr error
 }
 
 type fakeArchiveCall struct {
@@ -555,6 +557,13 @@ func (f *fakeWA) UploadNewsletter(ctx context.Context, data []byte, mediaType wh
 
 func (f *fakeWA) SendChatPresence(ctx context.Context, jid types.JID, state types.ChatPresence, media types.ChatPresenceMedia) error {
 	return nil
+}
+
+func (f *fakeWA) SendPresence(ctx context.Context, presence types.Presence) error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	f.presenceCalls = append(f.presenceCalls, presence)
+	return f.sendPresenceErr
 }
 
 func (f *fakeWA) DecryptReaction(ctx context.Context, reaction *events.Message) (*waProto.ReactionMessage, error) {
--- a/internal/app/heartbeat_test.go
+++ b/internal/app/heartbeat_test.go
@@ -32,6 +32,7 @@ func TestSyncWritesHeartbeatFileOnActivity(t *testing.T) {
 		func(string, string) {},
 		nil,
 		nil,
+		&syncPresence{},
 	)
 	defer f.RemoveEventHandler(handlerID)
 
@@ -68,6 +69,7 @@ func TestSyncOnceDoesNotWriteHeartbeatFile(t *testing.T) {
 		func(string, string) {},
 		nil,
 		nil,
+		&syncPresence{},
 	)
 	defer f.RemoveEventHandler(handlerID)
 
@@ -96,6 +98,7 @@ func TestSyncFollowDoesNotWriteHeartbeatOnKeepAliveTimeout(t *testing.T) {
 		func(string, string) {},
 		nil,
 		nil,
+		&syncPresence{},
 	)
 	defer f.RemoveEventHandler(handlerID)
 
@@ -196,6 +199,7 @@ func TestSyncFollowDoesNotReconnectOnFreshKeepAliveTimeout(t *testing.T) {
 		func(string, string) {},
 		nil,
 		nil,
+		&syncPresence{},
 	)
 	defer f.RemoveEventHandler(handlerID)
 
@@ -246,6 +250,7 @@ func TestSyncFollowEmitsStaleEvent(t *testing.T) {
 		func(string, string) {},
 		nil,
 		nil,
+		&syncPresence{},
 	)
 	defer f.RemoveEventHandler(handlerID)
 
--- a/internal/app/sync.go
+++ b/internal/app/sync.go
@@ -142,13 +142,25 @@ func (a *App) Sync(ctx context.Context, opts SyncOptions) (SyncResult, error) {
 		defer stopWebhook()
 	}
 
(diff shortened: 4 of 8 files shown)


dovocoder:
Ready for review: commits squashed. The unavailable presence send now succeeds on SIGTERM exit — verified with a live WhatsApp account (no more `send_presence_failed` warnings after `timeout`-induced shutdown).

dovocoder:
@clawsweeper re-review

Addressed all three findings:
1. **[P2] Connect cancellation** — Added `DetachSocket` field to `ConnectOptions`. Only sync sets it (`connectForSync` + `ReconnectWithBackoff`). Other commands keep the default (`false`) so `--timeout` and signal cancellation still bound the connection.
2. **[P2] Presence ordering** — Removed `go` from available presence sends in `sync_events.go`. They are now synchronous, so a delayed available send cannot run after the final unavailable cleanup send.
3. **[P3] CHANGELOG** — Reverted all CHANGELOG.md edits. Release notes are in the PR body only.

Verified with a live WhatsApp account: `timeout 30s` SIGTERM → graceful stop → unavailable presence sent successfully (no `send_presence_failed` warning).

dovocoder:
@clawsweeper re-review

Addressed all three remaining P2 findings:

1. **[P2] Keep detached connect attempts cancellable** — The detached context now has a 30-second dial timeout via `time.AfterFunc`. The connect attempt is bounded; after the dial succeeds, the timer is stopped so the websocket stays open. Other commands keep the caller's context (default `DetachSocket: false`).

2. **[P2] Keep available presence within the sync lifetime** — The event handler is now removed (`RemoveEventHandler`) BEFORE the unavailable cleanup send, so no late `Connected`/`PushNameSetting` event can fire an available presence after cleanup.

3. **[P2] Bound the final unavailable presence send** — The cleanup send now uses `context.WithTimeout(context.Background(), 5*time.Second)` so a stalled WhatsApp write cannot block shutdown.

Full gate passes. Live verification confirms unavailable send still succeeds on `timeout 30s` SIGTERM.

dovocoder:
@clawsweeper re-review

Both P1 findings addressed:

1. **[P1] Auth bootstrap / reconnect cancellation** — `DetachSocket` is gated with `!opts.AllowQR && opts.PairPhoneNumber == ""` so auth/pairing never detaches. The detached dial uses `context.AfterFunc(ctx, dialCancel)` so the dial is cancelled if the caller's context fires (SIGTERM, --max-reconnect deadline). After the dial succeeds, the AfterFunc is unregistered so the websocket stays alive.

2. **[P1] Slow available-presence send blocking shutdown** — All presence sends (both available and unavailable) now use `sendPresenceBounded()` which applies a 5-second timeout via `context.WithTimeout(context.Background(), 5*time.Second)`. A stalled WhatsApp write cannot hold the event handler dispatch lock or block the final unavailable cleanup send.

Full gate passes. Live verification: clean `timeout 30s` cycle → graceful stop → unavailable sent successfully, no warnings.

dovocoder:
@clawsweeper re-review

Both P1 findings addressed:

1. **[P1] Unavailable on ALL post-connect exits** — Moved the unavailable cleanup to a `defer` placed right after `connectForSync` succeeds. It now runs on every exit path (success, AfterConnect error, storage limit, reconnect failure), not just the success path. The manual cleanup at the end of `Sync()` was removed.

2. **[P1] Regression test for failure/limit exits** — Added `TestSyncSendsUnavailableOnErrorExit` which sets `AfterConnect` to return an error and verifies that `PresenceUnavailable` is still sent via the defer. The test confirms the call sequence `[available, unavailable]` — available on connect, unavailable on error exit.

Full gate passes. Live verification: clean `timeout 30s` cycle → graceful stop → unavailable sent successfully, no warnings.

dinakars777:
@dovocoder thanks!

PR #291, opened 2026-07-05
feat(media): backfill already-synced downloads

njt:
## Summary

Add `wacli media backfill` for media metadata already present in the local store but missing a local file. The command supports optional chat filtering, newest-first limits, bounded worker concurrency, human/JSON result counts, interruption, and an explicit opt-in timeout.

This maintainer rewrite narrows the original combined PR to backfill only:

- the one-shot drain bug already landed separately in #295
- the media-retry protocol and schema migration are excluded for separate review
- contributor credit is preserved in the commit and changelog

## Maintainer hardening

- cancellation reaches pending-media count/list queries and workers
- pre-canceled contexts cannot dispatch downloads
- unlimited backfill has no implicit five-minute deadline; explicit `--timeout` still works
- worker count is bounded by selected jobs
- newest-first ties use stable row order
- negative limits and worker counts are rejected
- sqlc output regenerated from the checked-in query

## Verification

Exact head: `33a1ecc`

- structured autoreview: clean after fixing cancellation and default-timeout findings
- `go test -race ./...`
- `pnpm format:check`
- `pnpm lint`
- `pnpm test`
- `pnpm build`
- `pnpm docs:site`
- `git diff --check`

Source-blind live validation on the exact built patch:

- sent the public repository README to the linked account's self-chat
- `media backfill --limit 1 --workers 1 --json` reported pending `2044`, attempted `1`, downloaded `1`, skipped `0`, failed `0`
- the downloaded fixture had a persisted local path
- downloaded bytes matched the sent fixture exactly by SHA-256
- negative-limit validation failed before store access

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+- Media: add `media backfill` to download media for already-synced messages that have metadata but no local copy. (thanks @njt)
+
 ### Security
 
 - Store: replace dynamic schema-inspection SQL with fixed allowlisted queries. (thanks @dovocoder)
--- a/cmd/wacli/media.go
+++ b/cmd/wacli/media.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/openclaw/wacli/internal/app"
 	"github.com/openclaw/wacli/internal/out"
 	"github.com/openclaw/wacli/internal/wa"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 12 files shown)


njt:
Added a second, related commit (`592e196`): **fix(sync): drain queued media before once/bootstrap exit**.

While testing backfill against a real hourly `sync --once --download-media` cron, I found `--once`/bootstrap cancels the media worker context at idle-exit, aborting downloads still queued or in flight. So a one-shot sync that receives a batch of media exits after idle and drops most of it — exactly the gap that then requires backfill (and only works within WhatsApp's ~30-day CDN window).

The fix adds a mutex-guarded `mediaQueue` that's safe to close under concurrent enqueue, splits the worker teardown into `wait`/`cancel`, and drains queued downloads on the clean once/bootstrap path before exit. Follow mode and error/limit/cancel paths keep the immediate hard-stop. New regression test `TestSyncOnceDrainsMediaBeforeExit` (downloads sized above the 250ms idle poll) fails without the drain and passes with it. `go test -race` clean.

Happy to split this into its own PR if you'd prefer to review the two independently.

njt:
Added a third commit (`399f780`): **feat(media): recover CDN-expired media via media-retry protocol**.

### Why
`media download`/`backfill` fetch from WhatsApp's CDN, which only retains media ~30 days — older media returns HTTP 403 and was unrecoverable. But WhatsApp's multi-device **media-retry protocol** (what WhatsApp Web uses) lets a linked device ask the phone to re-upload expired media. whatsmeow already exposes `SendMediaRetryReceipt`; wacli now uses it.

### What
`media retry [--chat] [--before] [--limit] [--batch] [--wait]`:
- Sends retry receipts in batches (default 32) with a second attempt for non-responders.
- Downloads whatever the phone re-serves using the fresh direct path.
- Marks genuinely-gone media (`messages.media_unavailable_at`, migration 20) so later runs skip it; `no_response` stays pending (transient).
- `wa.DecryptMediaRetry` normalizes both "not found" signals (decrypted `NOT_FOUND` result and the unencrypted `ErrMediaNotAvailableOnPhone`).

### Real-world validation
Ran against a live account with a 964-message expired-media backlog spanning 2019→2026:

```
requested=964  recovered=801  not_on_phone=16  no_response=136  failed=11
```

**83% recovered in a single pass**, real JPEG/MP4 files on disk, only 16 genuinely gone from the phone. Recovery works far past the 30-day CDN window — the phone retained media across the whole history. `no_response` items remain pending for a follow-up run.

Tests cover not-on-phone marking + pending exclusion, no-response leaving rows pending with a second attempt, and multi-batch processing. `go test -race` + vet clean.

This makes the PR a cohesive "media capture completeness" set: **backfill** (existing rows sync missed) + **once-mode drain fix** (don't cancel downloads at idle) + **retry** (recover CDN-expired media). Happy to split into separate PRs if you'd prefer.

steipete:
Thanks @njt — I isolated the reproducible one-shot media-drain bug into #295 and preserved your co-author credit. The fix now also fences active event producers, avoiding the queue-close and handler-unregistration races found during maintainer autoreview.

#291 remains open for the backfill, media-retry, and migration/product surface. #295 is draft only because the configured linked device became unauthenticated during live validation; local regression, race, full-gate, and structured review proof are green.

steipete:
The isolated one-shot media-drain fix is now merged in #295 as 54d436db0168edfa7b44f14f28370c39c162e75b. Live validation covered an authenticated connection, clean one-shot idle exit, and a real server media download with an exact SHA-256 match; the deterministic regression and race tests cover the accepted-job drain timing. Your co-author credit and changelog thanks are preserved. Thank you again, @njt.

Keeping this PR open for the broader backfill, retry, and schema changes, which need separate product review.

PR #304, opened 2026-07-15
fix: open SQLite stores on Windows

### LINKED ISSUE #303: `invalid uri authority` error on Windows upon `wacli auth`
Hello, I tried using `wacli` 0.12.0 for the first time on my Windows 10 computer, and got the following error: 
```shell
> wacli auth
create schema_migrations table: invalid uri authority: C:%5CUsers%5Cme%5C.wacli%5Cwacli.db?_foreign_keys=on&_busy_timeout=5000
```

According to Claude, this could be a fix: 
```go
// internal/sqliteutil/files.go
func FileURI(path, rawQuery string) string {
	p := filepath.ToSlash(path)
	if len(p) > 1 && p[1] == ':' { // Windows: C:/Users/... -> /C:/Users/...
		p = "/" + p
	}
	return (&url.URL{Scheme: "file", Path: p, RawQuery: rawQuery}).String()
}
```

Instead of the current:
```go
// internal/sqliteutil/files.go
func FileURI(path, rawQuery string) string {
	return (&url.URL{Scheme: "file", Path: path, RawQuery: rawQuery}).String()
}
```
(as well as i `internal/app/session_resolver.go`'s `readOnlySessionURI` and `cmd/wacli/auth_status_readonly.go`'s `fmt.Sprintf("file:%s?%s", …)`.

But I'm not knowledgeable enough, so I defer to you.

goutamadwant:
## Summary

- Normalize absolute Windows drive paths into valid SQLite file URIs instead of URI authorities.
- Preserve UNC paths with an empty URI authority and route writable, read-only session, and auth-status database openers through the same helper.
- Add regressions for drive-letter, UNC, and both read-only call paths while preserving Unix path escaping and query behavior.

Disclosure: AI was used to understand the codebase and review the fix.

## Why

On Windows, `wacli auth` could serialize `C:\Users\...` as `file://C:%5CUsers...`. SQLite interprets `C:` as a URI authority and rejects the database before QR pairing with `invalid uri authority`. A shared constructor prevents the writable and read-only database paths from diverging.

Closes #303

Thanks @LeoDupont for the report and suggested affected call sites.

## Validation

Before the fix, the focused regression produced:

```text
FileURI("C:\\Users\\me\\My Store\\wacli.db") = "file://C:%5CUsers%5Cme%5CMy%20Store%5Cwacli.db?_foreign_keys=on"
```

After the fix, drive-letter and UNC inputs serialize as:

```text
file:///C:/Users/me/My%20Store/wacli.db?_foreign_keys=on
file:////server/share/My%20Store/wacli.db?_foreign_keys=on
```

Passed:

- `go test ./internal/sqliteutil ./internal/app ./cmd/wacli -count=1`
- `go test -tags sqlite_fts5 ./internal/sqliteutil ./internal/app ./cmd/wacli -count=1`
- `pnpm test`
- `pnpm lint`
- `pnpm build`
- `go fmt ./...`
- `GOOS=windows GOARCH=amd64 go test -c ./internal/sqliteutil -o "$TMPDIR/sqliteutil.test.exe"`
- `git diff --check`

A live `wacli auth` run on Windows was not performed because an authenticated Windows environment was not available. The source-level regression runs on every host, and the helper package was also cross-compiled for Windows amd64.

No flags, environment variables, configuration, schema, or dependency behavior changed.

### DIFF
--- a/cmd/wacli/auth_status_readonly.go
+++ b/cmd/wacli/auth_status_readonly.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/openclaw/wacli/internal/sqliteutil"
 	"go.mau.fi/whatsmeow/types"
 )
 
@@ -53,7 +54,7 @@ func readOnlySessionSQLiteURI(path string) string {
 	if !sqliteSessionSidecarsExist(path) {
 		params += "&immutable=1"
 	}
-	return fmt.Sprintf("file:%s?%s", path, params)
+	return sqliteutil.FileURI(path, params)
 }
 
 func sqliteSessionSidecarsExist(path string) bool {
--- a/cmd/wacli/auth_test.go
+++ b/cmd/wacli/auth_test.go
@@ -134,6 +134,16 @@ func TestReadOnlyAuthStatusReadsLiveWALSidecars(t *testing.T) {
 	}
 }
 
+func TestReadOnlySessionSQLiteURIWindowsPath(t *testing.T) {
+	uri := readOnlySessionSQLiteURI(`C:\Users\me\.wacli\session.db`)
+	if !strings.HasPrefix(uri, "file:///C:/Users/me/.wacli/session.db?") {
+		t.Fatalf("readOnlySessionSQLiteURI = %q, want absolute Windows file URI", uri)
+	}
+	if !strings.Contains(uri, "mode=ro") {
+		t.Fatalf("readOnlySessionSQLiteURI = %q, want read-only query parameter", uri)
+	}
+}
+
 func TestReadOnlySessionSQLiteURIUsesLockingForLiveState(t *testing.T) {
 	path := filepath.Join(t.TempDir(), "session.db")
 	if err := os.WriteFile(path, []byte(""), 0o600); err != nil {
--- a/internal/app/app_test.go
+++ b/internal/app/app_test.go
@@ -149,6 +149,16 @@ func TestReadOnlySessionURIEscapesPathDelimiters(t *testing.T) {
 	}
 }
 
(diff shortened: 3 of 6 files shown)


steipete:
Maintainer proof on exact head `9d5b276b246dac84dcf690b3ffb8fa6570b4d0f6`:

- Local full gate passed: `pnpm format:check && pnpm lint && pnpm test && pnpm build && git diff --check`. This covered plain and `sqlite_fts5` Go suites, the Windows lock cross-compile, CGO guard, docs/release workflow tests, vet, and the real CLI build.
- Exact-head Windows artifact cross-build passed on AWS Crabbox `cbx_8cc6b8b3667c` (`run_0dbb2947a34f`) with `CGO_ENABLED=1 GOOS=windows GOARCH=amd64 CC=x86_64-w64-mingw32-gcc go build -tags sqlite_fts5`. SHA-256: `3bf8b2d10b90b5156be0b840783cbba81909c958692e5960f4c8a6e1453f039f`.
- The same artifact then passed a native Windows end-to-end run on AWS Crabbox `cbx_6e70a014f599` (`run_b380d6133fc7`): a fresh drive-letter store created `wacli.db`, `wacli auth --qr-format text` reached QR pairing without `invalid uri authority`, and `--read-only auth status` completed successfully. QR payload was not logged. Both leases were released.
- AutoReview passed with `.agents/skills/autoreview/scripts/autoreview --mode branch --base origin/main`: `autoreview clean: no accepted/actionable findings reported`.

The reported Windows drive-letter failure now has real runtime proof. UNC behavior remains source/regression-tested rather than live network-share-tested; that is outside the reporter's failing path.

PR #315, opened 2026-07-26
feat(sync): forward receipts and chat presence over the webhook

Jaime-data:
## Problem

The sync webhook only forwards messages. Delivery/read receipts and typing
notifications never leave the runtime, so a consumer building a WhatsApp-like UI
on top of `wacli sync --follow` cannot show message ticks or "typing…" at all —
the data is there in `whatsmeow`, it just has no way out.

## What this adds

`--webhook-events`, a comma-separated opt-in list (`message`, `receipt`,
`chat_presence`) that defaults to `message`. With the default, the payload stream
is what it has always been, plus one additive field.

Every webhook body now carries a flat `EventType` discriminator:

```json
{"EventType":"message","Chat":"…","ID":"3EB0…","Text":"hi","ChatName":"Alice"}
{"EventType":"receipt","Chat":"…@g.us","Sender":"…@s.whatsapp.net","MessageIDs":["3EB0…"],"Timestamp":"…","Type":"delivered","IsFromMe":false}
{"EventType":"chat_presence","Chat":"…","Sender":"…","State":"composing","Media":""}
```

Flat and additive rather than a `{type, data}` envelope on purpose: re-nesting the
message would break every existing consumer, and the discriminator alone is enough
to switch on.

### Design notes

- **Receipts are filtered at the source** to `delivered`, `read` and `played`. The
  bookkeeping types (`sender`, `retry`, `read-self`, `played-self`, `inactive`,
  `server-error`, `peer_msg`, `hist_sync`) are dropped before enqueuing: in groups
  they multiply by participant and would compete with real messages for the bounded
  queue. Happy to reverse this and forward `Type` verbatim if you would rather the
  webhook not have an opinion.
- **`delivered` is spelled out**, even though WhatsApp sends it as an empty string.
  A consumer switching on `"delivered"` silently never matches otherwise.
- `MessageIDs` keeps WhatsApp's batching (one POST per receipt, not per message),
  and in groups `Sender` is the participant the receipt came from.
- **`events.Presence` (online / last seen) is not forwarded** — out of scope here,
  and it would need `SubscribePresence` anyway.
- **The queue stays at 512 with one worker.** The full-queue warning now reports the
  dropped event kind and a running total: a dropped message is recoverable by
  reconciliation, a dropped `delivered` is a tick that never advances and today
  leaves no trace.
- `chat_presence` only produces events under `--presence-mode normal`, since
  WhatsApp only sends typing notifications to devices that mark themselves
  available. Documented on the flag.

## Docs and tests

`docs/sync.md` documents the flag and all three payload shapes, next to the
existing webhook docs. Tests cover flag parsing and validation, the receipt-type
filter and the empty-string translation, batch preservation, the JSON shape of
each payload, that the default forwards messages only, that the opt-in delivers
all three kinds, and that `events.Presence` never reaches the webhook.

`pnpm test`, `pnpm lint`, `pnpm format:check` and the `deadcode` gate are green
locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.15.1 - Unreleased
 
+### Added
+
+- Sync: add opt-in receipt and chat-presence webhook events while keeping legacy message payloads unchanged. (#315 - thanks @Jaime-data)
+
 ### Chore
 
 - Build: migrate `sqlc` code generator to Go 1.24+ `go tool` directive and bump project toolchain requirement to Go 1.26.5. (#313 - thanks @thedavidweng)
--- a/cmd/wacli/sync.go
+++ b/cmd/wacli/sync.go
@@ -25,6 +25,7 @@ func newSyncCmd(flags *rootFlags) *cobra.Command {
 	var webhookURL string
 	var webhookSecret string
 	var webhookAllowPrivate bool
+	var webhookEventsFlag string
 	var storage syncStorageLimitFlags
 
 	cmd := &cobra.Command{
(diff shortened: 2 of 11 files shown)


steipete:
Maintainer rewrite pushed for compatibility before landing.

- Legacy message webhook bodies remain byte-for-byte unchanged and omit `EventType`; its absence means `message`.
- Only the new `receipt` and `chat_presence` payloads carry `EventType`, and both now have permanent HMAC coverage against the exact transmitted body.
- The branch is rebased onto current `main`. Jaime retains authorship of the implementation commit, while the changelog is a separate maintainer commit thanking @Jaime-data.

The full local gate and final Codex autoreview are clean on `d309636`.

PR #318, opened 2026-07-27
feat: add opt-in --send-spacing to pace delegated sends

cohnen:
## What

Adds an opt-in `sync --follow --send-spacing <min>[-<max>]` flag that spaces the daemon's serialized delegated sends.

- `--send-spacing 2s` — fixed 2s minimum gap between sends.
- `--send-spacing 500ms-5s` — a fresh random gap per send drawn from the range (jitter).
- Unset (default) — behavior is unchanged (the existing warn-only rapid-send guardrail; no delay).

## Why

When an automation fires several `wacli send` processes at a running `sync --follow` daemon, the daemon already serializes them under `sendMu` in `handleSendDelegateConn` — but with **zero gap**, so the burst still leaves the wire back-to-back. WhatsApp punishes bursts from one account with rate-limiting, decrypt-retry storms, and account restrictions. The current guardrail (`warnRapidSendIfNeeded`) only prints a warning; nothing actually paces the sends.

This adds pacing at the one place every delegated send funnels through, so it covers all delegated send kinds (text, file, voice, sticker, react, poll, select) without touching each command.

## How it works

- `send_spacing.go`
  - `sendSpacing` + `parseSendSpacing` — parses `""` (disabled), a single duration (fixed gap), or a `min-max` range (validated: non-negative, `min <= max`).
  - `sendPacer` — `wait(ctx)` blocks out the remaining gap since the previous send start and reports whether the caller may proceed; `record()` advances the baseline only once a send is actually dispatched. Clock, sleeper, and RNG are injectable for tests; `sleepWithContext` returns early on ctx cancellation.
- `send_ipc.go` — one shared `sendPacer` per delegate server. Inside the existing `sendMu` critical section, pacing runs under a context **bounded by the caller's `req.TimeoutMS`**, and pacing + send share that one deadline. If the wait can't complete within the budget, the send is **aborted before dispatch** (returns a timeout error) so a long spacing interval can never deliver after the caller has already timed out. Disabled spacing leaves the path byte-for-byte unchanged.
- `sync.go` — the `--send-spacing` flag, parsed early in `RunE` so a bad value fails fast.

Scope note: this covers the **daemon/delegated** path (where automated fan-out funnels and the burst problem actually shows up). The direct path — multiple standalone `wacli send` processes with no daemon running — is already serialized by the store LOCK; a symmetric `--send-spacing` on the `send` subcommands (using the existing `.last-send-at` marker) would be a natural, small follow-up if wanted.

## New flags / env

- `sync --follow --send-spacing <min>[-<max>]` (e.g. `2s`, `500ms-5s`). No new env vars, no new dependencies.

## Testing

- `send_spacing_test.go`: table-driven `parseSendSpacing` cases (fixed, range, spaces, equal bounds, `max<min`, unparseable, negative) and `sendPacer` behavior (disabled no-op, first-send no wait, fixed and random gaps within bounds, elapsed-exceeds-gap skip, timeout-aborts-before-dispatch, already-expired-context blocks first send).
- Full gate green: `gofmt -l` clean, `go vet ./...`, `go build -tags sqlite_fts5` (CGO), `go test ./...` and `go test -tags sqlite_fts5 ./...`, windows-lock cross-compile, and the `cgo-required` guard — all pass; `git diff --check` clean.

## Live behavior proof

Ran against a **real linked WhatsApp account** on a clean VM (arm64, glibc 2.43) with a Go 1.26.5 CGO build of this branch. Daemon: `wacli sync --follow --send-spacing 2s-5s`. Every send targets the linked account's own "message yourself" chat — no third party. Numbers, JIDs, and names are redacted.

### 1. Spacing — five delegated sends fired in parallel

All five `wacli send` processes launched at the same instant; each fails the store lock and delegates to the running daemon. Client-side completion order/timing:

```
send #1 completed at +0.00s (first)
send #4 completed at +2.26s total, +2.26s after previous
send #2 completed at +5.00s total, +2.73s after previous
send #3 completed at +7.80s total, +2.81s after previous
send #5 completed at +10.50s total, +2.69s after previous
```

Consecutive gaps **2.26 / 2.73 / 2.81 / 2.69s** — every one inside the configured `[2s, 5s]`, with visible jitter. Without pacing all five complete within ~1s. A delegated result (redacted): `{"success":true,"data":{"id":"3EB0…","sent":true,"to":"<redacted>@lid"}}`.

### 2. Timeout safety (the P1 fix) — no dispatch after the caller's timeout

- **A** — normal send → `sent:true`; freshens the daemon's last-send time.
- **B** — `--timeout 300ms` → the remaining spacing gap (≥ ~1s) can't be met within 300ms. B fails client-side at **0.31s**: `read unix …/.send.sock: i/o timeout`.
- **9 seconds later** (well past the 5s max gap), a full-text message search for the demo texts returns **only A**:

```
timeout-demo-A-normal        (FromMe, sent)
timeout-demo-B-shorttimeout  — ABSENT (never dispatched)
```

B was never put on the wire. Under the pre-fix code the pacer waited under the daemon's own long-lived context and would have dispatched B ~2–5s after the caller already reported failure; the fix bounds pacing by the request deadline and skips dispatch when the budget is exhausted.

_(Session logged out / device unlinked and the VM store wiped after capture.)_

### DIFF
--- a/cmd/wacli/send_ipc.go
+++ b/cmd/wacli/send_ipc.go
@@ -140,6 +140,11 @@ func startSendDelegateServer(ctx context.Context, a *app.App, spacing sendSpacin
 
 	done := make(chan struct{})
 	var sendMu sync.Mutex
+	var pacedSendSlot chan struct{}
+	if spacing.enabled() {
+		pacedSendSlot = make(chan struct{}, 1)
+		pacedSendSlot <- struct{}{}
+	}
 	// One pacer shared across connections: it spaces the serialized delegated
 	// sends so a burst of `wacli send` processes delegating to this daemon
 	// leaves the wire paced instead of back-to-back. Disabled = no-op.
@@ -151,7 +156,7 @@ func startSendDelegateServer(ctx context.Context, a *app.App, spacing sendSpacin
 			if err != nil {
 				return
 			}
-			go handleSendDelegateConn(ctx, conn, a, &sendMu, pacer)
+			go handleSendDelegateConn(ctx, conn, a, &sendMu, pacedSendSlot, pacer)
 		}
 	}()
 
@@ -177,7 +182,7 @@ func removeStaleSendDelegateSocket(path string) error {
 	return os.Remove(path)
 }
 
-func handleSendDelegateConn(ctx context.Context, conn net.Conn, a *app.App, sendMu *sync.Mutex, pacer *sendPacer) {
+func handleSendDelegateConn(ctx context.Context, conn net.Conn, a *app.App, sendMu *sync.Mutex, pacedSendSlot chan struct{}, pacer *sendPacer) {
 	defer conn.Close()
 	_ = conn.SetDeadline(time.Now().Add(5 * time.Minute))
 
@@ -193,8 +198,23 @@ func handleSendDelegateConn(ctx context.Context, conn net.Conn, a *app.App, send
 		defer cancel()
 	}
 
-	sendMu.Lock()
-	defer sendMu.Unlock()
+	if pacer.enabled() {
+		select {
+		case <-requestCtx.Done():
+			_ = json.NewEncoder(conn).Encode(sendDelegateResponse{
+				OK:    false,
+				Error: "send spacing exceeded request timeout before dispatch",
+			})
+			return
+		case <-pacedSendSlot:
+			defer func() { pacedSendSlot <- struct{}{} }()
+		}
+	} else {
+		// Preserve the original unpaced serialization path exactly when the
+		// opt-in flag is unset.
+		sendMu.Lock()
+		defer sendMu.Unlock()
+	}
 
 	// Space this send from the previous one while serialized. Bound the wait by
 	// the caller's request timeout, including time spent waiting for earlier
--- a/cmd/wacli/send_ipc.go
+++ b/cmd/wacli/send_ipc.go
@@ -59,6 +59,7 @@ type sendDelegateRequest struct {
 	PresenceMedia        string   `json:"presence_media,omitempty"`
 	PostSendWaitMS       int64    `json:"post_send_wait_ms,omitempty"`
 	TimeoutMS            int64    `json:"timeout_ms,omitempty"`
+	DeadlineUnixMS       int64    `json:"deadline_unix_ms,omitempty"`
 }
 
 type sendDelegateResponse struct {
@@ -96,7 +97,9 @@ func delegateSend(ctx context.Context, flags *rootFlags, req sendDelegateRequest
 	}
 	defer conn.Close()
 
-	_ = conn.SetDeadline(time.Now().Add(commandTimeout(flags)))
+	deadline := time.Now().Add(commandTimeout(flags))
+	req.DeadlineUnixMS = deadline.UnixMilli()
+	_ = conn.SetDeadline(deadline)
 	if err := json.NewEncoder(conn).Encode(req); err != nil {
 		return sendDelegateResponse{}, err
 	}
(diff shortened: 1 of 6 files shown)


cohnen:
Thanks for the review. Addressed the P1 in `6d0cdfc`.

### P1 — pacing could dispatch after the caller's timeout (fixed)

Root cause was exactly as flagged: `pace()` waited under the daemon's long-lived context, before `executeDelegatedSend` derived the request-scoped timeout.

Fix: split the pacer into `wait()` / `record()`. In `handleSendDelegateConn` the wait now runs under a context bounded by the caller's `req.TimeoutMS`, and pacing + send **share that one deadline**:

- If the wait can't complete within the budget, the send is **aborted before dispatch** and the caller gets `send spacing exceeded request timeout before dispatch` — no late send.
- `record()` advances the spacing baseline **only when a send is actually dispatched**, so an aborted send never shifts spacing.
- An already-expired context blocks even the first send.
- Disabled spacing (`--send-spacing` unset) leaves the path byte-for-byte unchanged.

### Behavior proof for the timeout path (deterministic)

Two regression tests drive the exact failure mode with an injected clock/sleeper (no live account needed):

- `TestSendPacerAbortsWhenTimeoutElapsesDuringWait` — the wait is cut short by the request timeout → send is **not** dispatched and the baseline is unchanged.
- `TestSendPacerFirstSendRespectsAlreadyExpiredContext` — an already-cancelled context blocks the first send.

Plus the existing spacing coverage (fixed gap, random gap within `[min,max]`, elapsed-exceeds-gap skip, disabled no-op).

Full gate green on `go1.26.5`: `gofmt` clean, `go vet ./...`, `go build -tags sqlite_fts5` (CGO), `go test ./...` and `-tags sqlite_fts5 ./...`, windows-lock cross-compile, cgo guard, `git diff --check`.

### Live follow-daemon run

A redacted terminal capture of a live `sync --follow` spacing real delegated sends is environment-gated for me right now — no linked test WhatsApp account is available in my setup, and I don't want to exercise a real personal number. The spacing + timeout-safety logic is fully covered by the unit tests above with an injected clock. If a maintainer can point me at a sanctioned test account (or confirm the unit-level proof is sufficient for a daemon-only, opt-in, default-off flag), I'll add the live capture.

### Product decision (P2 — CLI surface)

`--send-spacing` is opt-in, default-off, and scoped to `sync --follow` (the delegated-send funnel). The direct multi-process path is already LOCK-serialized; a symmetric flag on the `send` subcommands would be a small follow-up if wanted. Happy to adjust naming/placement per your preference.

@clawsweeper re-review

cohnen:
Added the **Live behavior proof** section to the PR description — ran this branch against a real linked WhatsApp account (sends to the account's own "message yourself" chat, everything redacted):

1. **Spacing** — five delegated sends fired in parallel completed **+2.26 / +2.73 / +2.81 / +2.69s** apart, every gap inside the configured `[2s, 5s]`.
2. **Timeout safety (the P1)** — a send with `--timeout 300ms` whose spacing gap couldn't be met failed at 0.31s and, checked 9s later (past the max gap), was **never delivered** — the daemon aborted before dispatch. Pre-fix it would have landed ~2–5s after the caller timed out.

Session logged out / device unlinked and the store wiped after capture.

@clawsweeper re-review

cohnen:
The review was success  https://github.com/openclaw/clawsweeper/actions/runs/30273892900 but no update here. Can you check again? @clawsweeper re-review Thanks a lot!

steipete:
Thanks @cohnen — this landed in `17d880997000a6d0acc4ec39b86180143d351387`.

Verification:

- `make check` passed four times during feature hardening and once more on merged `main` with the maintainer changelog entry. That covered formatting, vet/lint, source vulnerability policy, plain and FTS test suites, Windows cross-compilation, docs/release-policy tests, the build, GoReleaser config checks, and diff hygiene. The final `AGENTS_HOME="$HOME/.claude" AUTOREVIEW="$HOME/.claude/skills/autoreview/scripts/autoreview" "$HOME/.claude/skills/autoreview/scripts/autoreview" --mode branch --base origin/main` pass was clean, and GitHub's exact-head dispatch check passed.
- The pacing tests prove that unset spacing is a no-op; the first send is immediate; three fixed-gap sends produce two waits; randomized gaps remain within their configured bounds; elapsed idle time is credited; the gap starts after the prior attempt completes; timeout cancellation while queued or pacing prevents dispatch; the caller's absolute deadline is honored; long request budgets extend the IPC response deadline; and the full `time.Duration` range cannot panic the random picker.

Maintainer fixes applied on top:

- included mutex queue time, pacing, and execution in the caller's deadline
- made paced queue acquisition context-cancellable while preserving the original unpaced mutex path
- started each gap after the preceding attempt completes, so slow preparation cannot consume the intended wire spacing
- prevented RNG overflow at the maximum duration range
- propagated the caller's absolute deadline and kept the IPC transport alive through the final response
- clarified the flag help and added sync documentation

The maintainer-owned changelog entry landed separately in #321 at `f7e1ebeb3ffce620c479f23d6663e95d9c3157e6`.

PR #325, opened 2026-07-28
feat(sync): stop `sync --follow` on logout instead of reconnecting forever

cohnen:
## What
`sync --follow` now handles `*events.LoggedOut`: it emits a structured `logged_out` event and stops the loop cleanly, instead of reconnecting forever against a revoked session.

## Why
When WhatsApp revokes the session — the linked device is removed from the phone, or a logout/ban — whatsmeow emits `*events.LoggedOut`. wacli had no case for it, and whatsmeow keeps issuing `Disconnected` around the logout, so the follow loop's `<-disconnected → reconnect` path spun forever against a dead session. Nothing surfaced the logout and the daemon never stopped.

## What changed
- `internal/app/sync_events.go`: add `case *events.LoggedOut` → `emitOrPrint("logged_out", {reason, reason_code, on_connect}, …)` and signal a new buffered `loggedOut` channel (mirrors the existing `disconnected` plumbing).
- `internal/app/sync_idle.go`: both `runSyncFollow` and `runSyncUntilIdle` gain a `case <-loggedOut` that emits a terminal `stopping` line (`reason: logged_out`) and returns `nil` — clean stop, no reconnect.
- `internal/app/sync.go`: create the `loggedOut` channel and thread it through.
- **Review follow-up (`f322470`):** a revocation delivers `Disconnected` and `LoggedOut` back to back, and with both buffered channels ready `select` picks at random — so the loop could start one reconnect against the dead session before consuming the terminal signal. Every reconnect path (`disconnected` + `staleReconnect` in follow mode, `disconnected` in idle mode) now drains `loggedOut` non-blockingly first, so the terminal logout signal always wins.
- **Review follow-up (`333b27c`):** the `logged_out` output now names the recovery sequence in both modes — a `recovery` field on the NDJSON event and a printed hint on the human path: run `wacli auth logout` to clear the local session, then `wacli auth --phone` to pair again. Regression tests assert the wording in both the `--events` data and the human status-line output.

**Scope note:** this reports and stops only — it deliberately does **not** delete the local session store; that stays `auth logout`'s job. One consequence worth flagging: after a forced logout the `whatsmeow_device` row lingers, so a bare `auth --phone` re-pair short-circuits (it only pairs when `Store.ID == nil`) — which is why the emitted output now spells out the `auth logout` → `auth --phone` recovery sequence. If you'd prefer `sync` to also clear the session on logout (which would make `auth status` report not-authenticated and let re-pair start fresh), I'm happy to add it here or as a follow-up — I kept it out to keep this change non-destructive.

## New event
```json
{"event":"logged_out","data":{"reason":"401: logged out from another device","reason_code":401,"on_connect":true,"recovery":"To re-authenticate, run `wacli auth logout` to clear the local session, then `wacli auth --phone` to pair again."},"ts":…}
```

## Tested
Deterministic fake-client unit tests (no live account), in `internal/app/sync_logout_test.go`:
- the handler emits `logged_out` and signals the channel — both the forced-logout (`OnConnect=true`, reason 401) and stream-error (`OnConnect=false`, reason 0) paths;
- `runSyncFollow` and `runSyncUntilIdle` each stop on the logout signal with **no reconnect** (`connectCalls == 0`), under a background context so only the logout — not ctx cancellation — ends the loop;
- **paired-signal races:** `loggedOut` queued together with `disconnected` (and, in follow mode, with a `staleReconnect` request) still stops with `connectCalls == 0` — 50 iterations per case to cover `select`'s random choice between ready channels. These fail on the pre-fix head `d7be620` within a few runs;
- end-to-end `sync --follow` stops (`Sync` returns nil, `connectCalls == 1`) when a `LoggedOut` is delivered after connect;
- **recovery-hint wording:** the `logged_out` NDJSON `recovery` field and the human status-line output must both name `wacli auth logout` and `wacli auth --phone`.

`make check` is green locally on go1.26.5: gofmt, `go vet`, `go test ./...` plain + `-tags sqlite_fts5`, windows-lock cross-compile, cgo-required guard, docs tests, build, GoReleaser config checks, and `git diff --check`.

## Real behavior proof (live linked account)

Binary built from this PR's head (`f322470`, `CGO_ENABLED=1 -tags sqlite_fts5`, linux/arm64), paired to a real WhatsApp account via `auth --phone` (pair code) in a throwaway store; bootstrap synced ~3.9k messages. Then `sync --follow --events` ran while the linked device was removed from the phone (**Settings → Linked devices → Log out**). Phone numbers/JIDs never appear in the NDJSON; local IPs/ports redacted below.

```console
$ wacli --store /tmp/proof-store --events sync --follow
{"event":"connected","ts":1785339889247}

# ~35 min of normal operation. Several transient network drops during the run
# each took the normal path — disconnected → reconnecting → connected — e.g.:
16:19:12.623 [Client/Socket ERROR] Error reading from websocket: failed to get reader: failed to read frame header: read tcp REDACTED->REDACTED:443: read: connection reset by peer
{"event":"disconnected","ts":1785341952623}
{"event":"reconnecting","ts":1785341952623}
{"event":"connected","ts":1785341953614}

# 16:35 UTC: linked device removed from the phone →
{"event":"logged_out","data":{"on_connect":false,"reason":"401: logged out from another device","reason_code":401},"ts":1785342925277}
{"event":"stopping","data":{"messages_synced":1,"reason":"logged_out"},"ts":1785342925277}
Messages stored: 1
$ # process exited 0 — no reconnecting/connected events after logged_out
```

The contrast is the point: ordinary disconnects still reconnect (the pre-existing path, exercised repeatedly in the same run), while the real revocation emitted `logged_out`, stopped follow mode, and made no further reconnect attempt.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Fixed
 
+- Sync: stop promptly when WhatsApp revokes the linked session, including while a reconnect is already in progress, and print the re-authentication steps. (#325 - thanks @cohnen)
 - Send: keep self-chat storage under the canonical phone-number chat and reject text sends to the linked account itself instead of reporting an acknowledgement that may never reach Message Yourself. (#319 - thanks @Lucas-Kim-J)
 
 ### Chore
--- a/docs/sync.md
+++ b/docs/sync.md
@@ -16,6 +16,7 @@ wacli sync [--once] [--follow] [--idle-exit 30s] [--max-reconnect 5m] [--stale-t
 - `--once` exits after sync becomes idle.
 - `--idle-exit` controls idle exit timing in once mode.
 - `--max-reconnect 0` keeps reconnecting indefinitely.
+- If WhatsApp revokes the linked session, sync emits a terminal `logged_out` event, cancels any reconnect already in progress, and exits cleanly. Re-pair with `wacli auth logout` followed by `wacli auth --phone`.
 - `--max-messages N` stops before storing more than `N` total messages locally.
 - `--max-db-size SIZE` stops when `wacli.db` plus SQLite sidecars reaches `SIZE` (`500MB`, `2GB`, etc.).
 - `--download-media` runs a bounded media downloader for sync events. Clean one-shot and bootstrap runs finish queued downloads before exiting; cancellation, errors, and storage-limit exits stop immediately.
--- a/internal/app/fake_wa_test.go
+++ b/internal/app/fake_wa_test.go
@@ -34,12 +34,13 @@ type fakeWA struct {
 	nextHandlerID uint32
 	handlers      map[uint32]func(interface{})
 
-	connectEvents []interface{}
-	connectErrs   []error
-	connectCalls  int
-	connectDelay  time.Duration
-	downloadDelay time.Duration
-	downloadErr   error
+	connectEvents  []interface{}
+	connectErrs    []error
+	connectCalls   int
+	connectDelay   time.Duration
+	connectStarted chan struct{}
+	downloadDelay  time.Duration
+	downloadErr    error
 
 	onMediaRetry       func(info *types.MessageInfo, mediaKey []byte) interface{}
 	mediaRetryReceipts []string
@@ -178,6 +179,12 @@ func (f *fakeWA) Connect(ctx context.Context, opts wa.ConnectOptions) error {
 		return nil
 	}
 	f.connectCalls++
+	if f.connectStarted != nil {
+		select {
+		case f.connectStarted <- struct{}{}:
+		default:
+		}
+	}
 	authed := f.authed
 	var connectErr error
 	if len(f.connectErrs) > 0 {
--- a/internal/app/heartbeat_test.go
+++ b/internal/app/heartbeat_test.go
@@ -28,6 +28,7 @@ func TestSyncWritesHeartbeatFileOnActivity(t *testing.T) {
 		&messagesStored,
 		&lastEvent,
 		make(chan struct{}, 1),
+		make(chan struct{}, 1),
 		make(chan staleReconnectRequest, 1),
 		func(string, string) {},
 		nil,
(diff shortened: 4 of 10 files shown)


cohnen:
@clawsweeper re-review

Pushed `333b27c` addressing the remaining P2: the `logged_out` output now names the verified recovery sequence (`wacli auth logout` → `wacli auth --phone`) in both the human output and the NDJSON `recovery` field, with regression tests asserting the wording on both paths.

steipete:
Maintainer follow-up pushed in 2234205.

I reproduced the remaining race before the fix by blocking `ReconnectWithBackoff`, delivering `LoggedOut` after reconnect had started, and observing `runSyncFollow` remain stuck past the two-second assertion. The new reconnect wrapper owns a derived context, cancels it when the terminal logout signal arrives, and treats logout as the terminal result even if reconnect completes concurrently. The same regression now covers follow and idle modes.

Proof on the pushed head:

- `go test ./internal/app -run 'LoggedOut' -count=1 -v` — all logout handler, queued-signal, in-flight reconnect, recovery-output, and end-to-end fake-client cases pass.
- `pnpm format:check && pnpm lint && pnpm test` — green, including plain Go, FTS5, Windows lock cross-compile, CGO guard, and documentation/release tests.
- `pnpm build && pnpm docs:site && git diff --check` — green.
- Built CLI: `./dist/wacli --version` prints `wacli 0.15.0`; `./dist/wacli sync --help` exposes `--max-reconnect` and global `--events`.
- Source-blind behavior probes against two independent empty stores each emitted `{"event":"error","data":{"message":"not authenticated; run `wacli auth`"},...}` and exited 1; an invalid reconnect duration was rejected before store/network work.
- Codex autoreview: clean, no accepted/actionable findings (overall correctness 0.96).

The earlier contributor trace remains the real-account revocation proof. I did not revoke a personal linked-device session merely to repeat it; the current-head delta is the deterministic cancellation regression plus docs/changelog.

@clawsweeper re-review

PR #326, opened 2026-08-01
fix(send): resolve quoted messages under the chat's PN/LID alias

0xlucuma:
## Summary

- fall back to the chat's PN/LID counterpart when looking up a quoted message
- resolve the counterpart only when a reply is requested, so plain sends are unchanged
- normalize the resolved JID with `ToNonAD()`, matching `canonicalMessageFilterJID`

## Why

`send text --reply-to` fails in one-to-one chats with `quoted message <id> not found in local store for chat <lid>; run `wacli sync` first`, and syncing never helps.

Sends resolve the recipient to whichever addressing mode WhatsApp reports, so a LID-addressed contact becomes `<lid>@lid`, and `buildTextReplyContextInfo` does a single exact-match `GetMessage` against it. The store canonicalizes the other way, because the LID migration on authenticated startup re-homes `@lid` rows to the phone-number JID. The lookup therefore searches the LID chat while the history sits under the PN chat. On a live store that was 181 messages under `<pn>` against 1 under `<lid>`. Reads already compensate through `mappedChatJIDs`, but the send path does not. Groups escape it through the `ErrNoRows` plus `--reply-to-sender` fallback, so only DMs are affected.

## Validation

Full repository gate green. New table-driven regression test `TestBuildTextReplyContextInfoFindsQuoteUnderChatAlias` covers both directions and asserts the un-aliased lookup still fails, so it cannot pass vacuously. Verified against a live account, where a reply that previously failed was delivered and quoted correctly.

## Redacted live behaviour proof

`textReplySelfJID` ran its own unaliased lookup before the alias existed, so an
own quoted message under the counterpart JID produced the linked phone JID as
`ContextInfo.Participant` instead of the linked LID. Addressed in the second
commit by resolving the alias first and passing it into that lookup, plus a
send-path regression that goes through `sendTextMessageWithSender` with a
`FromMe` quote.

Proof below uses a temporary diagnostic build with stderr-only tracing that
stops before the wire send. The tracing is not part of the PR diff. It runs
against a scratch copy of an authenticated store, put into the split state the
bug requires: 188 messages under the phone JID, 1 under the LID, both chats
present. Numbers redacted.

```text
### main
$ WACLI_PROOF=1 wacli-proof-stock --store REDACTED --json send text \
    --to <redacted>@s.whatsapp.net --reply-to 3EB0B3DD... --message proof
{"success":false,"data":null,"error":"quoted message 3EB0B3DD... not found in
 local store for chat <redacted>@lid; run `wacli sync` first"}

### first commit only (quote lookup aliased, participant lookup not)
proof: chat=<redacted>@lid alias=<redacted>@s.whatsapp.net
       selfJID=<redacted>@s.whatsapp.net stanza=3EB0B3DD...
       participant=<redacted>@s.whatsapp.net      <-- linked PN, the reported bug

### this PR head
proof: chat=<redacted>@lid alias=<redacted>@s.whatsapp.net
       selfJID=<redacted>@lid stanza=3EB0B3DD...
       participant=<redacted>@lid                 <-- linked LID, matches main's
                                                      LID-addressed branch
```

Reverting only the ordering change makes the new regression fail with
`participant = "<linked pn>", want the linked LID`, so it cannot pass
vacuously.

End to end on a real linked account, before the second commit existed, a reply
that had failed with the `not found in local store` error above was delivered
and quoted correctly once the first commit was in place.

## Reproducing

The failure needs the split state, which is why it can look unreproducible: the
LID migration on authenticated startup merges the LID chat into the phone-JID
chat, and after that the lookup hits the right rows and stock succeeds. The bug
appears in the window between an `@lid`-addressed message creating a LID chat
and the next authenticated startup.

## Scope

Left out to keep this reviewable, happy to follow up:

- the same single-JID lookup exists at other by-id call sites and would fail the same way: `send react` (`send_react_cmd.go:151`), `send select` (`send_select_cmd.go:224`), `media download` (`media.go:255`), `messages purge` (`messages_purge.go:46`), `messages_helpers.go:145`, and `poll vote/show` (`poll_cmd.go:258`, in a file that already resolves aliases for list filters)
- a shared helper in `internal/wa` would cover all of them, since it already owns `ResolvePNToLID` and `ResolveLIDToPN`, if you want one and can say where it should live
- wacli does not persist reply context for messages it sends, so an outbound reply is stored locally without its quote even though the quote goes out on the wire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@
 
 ### Fixed
 
-- Send: resolve quoted direct messages across phone-number and LID chat aliases so replies can find migrated history. (#326 - thanks @0xlucuma)
 - Send: keep self-chat storage under the canonical phone-number chat and reject text sends to the linked account itself instead of reporting an acknowledgement that may never reach Message Yourself. (#319 - thanks @Lucas-Kim-J)
+- Send: resolve quoted direct messages across phone-number and LID chat aliases so replies can find migrated history. (#326 - thanks @0xlucuma)
 
 ### Chore
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Fixed
 
+- Send: resolve quoted direct messages across phone-number and LID chat aliases so replies can find migrated history. (#326 - thanks @0xlucuma)
 - Send: keep self-chat storage under the canonical phone-number chat and reject text sends to the linked account itself instead of reporting an acknowledgement that may never reach Message Yourself. (#319 - thanks @Lucas-Kim-J)
 
 ### Chore
--- a/cmd/wacli/send.go
+++ b/cmd/wacli/send.go
@@ -218,6 +218,7 @@ type textMessageSender interface {
 	SendProtoMessage(ctx context.Context, to types.JID, msg *waProto.Message) (types.MessageID, error)
 	GetGroupInfo(ctx context.Context, jid types.JID) (*types.GroupInfo, error)
 	ResolvePNToLID(ctx context.Context, jid types.JID) types.JID
+	ResolveLIDToPN(ctx context.Context, jid types.JID) types.JID
 	LinkedJID() string
 	LinkedLID() string
 }
@@ -250,7 +251,11 @@ func sendTextMessageWithSender(ctx context.Context, sender textMessageSender, db
 	if err != nil {
 		return "", err
 	}
-	msg, plainText, err := buildTextMessageWithSelf(db, to, text, replyTo, replyToSender, selfJID, preview, mentionedJIDs)
+	var aliasTo types.JID
+	if strings.TrimSpace(replyTo) != "" {
+		aliasTo = replyLookupAlias(ctx, sender, to)
+	}
+	msg, plainText, err := buildTextMessageWithSelf(db, to, aliasTo, text, replyTo, replyToSender, selfJID, preview, mentionedJIDs)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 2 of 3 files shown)


0xlucuma:
Thanks, the P1 is correct and now fixed in 1d3c039.

`textReplySelfJID` ran its own exact `GetMessage` before `aliasTo` existed, so for a LID-addressed chat whose own quoted message sits under the counterpart JID it returned the linked PN and skipped the LID branch, and the aliased quote lookup then stamped that PN as the participant. The alias is now resolved before that call and passed into it, so both lookups agree.

Added `TestSendTextReplyToOwnMessageUnderChatAliasUsesLIDParticipant`, which goes through `sendTextMessageWithSender` with a `FromMe` quote and asserts the participant. Reverting only the ordering makes it fail with `participant = "<linked pn>", want the linked LID`, so it is not vacuous. Writing it also caught that my `ResolveLIDToPN` test double was an identity stub, which would have hidden alias behaviour in later tests.

The PR body now has a redacted three-way trace on a scratch copy of an authenticated store, using a temporary diagnostic build that stops before the wire send: main fails the lookup, the first commit produces the wrong participant, this head produces the linked LID.

One correction to my original description worth recording. The failure needs the store in the split state, and the LID migration on authenticated startup merges the LID chat into the phone-JID chat, after which stock succeeds. That is why it can look unreproducible. The window is between an `@lid`-addressed message creating a LID chat and the next authenticated startup, which is where I first hit it.

steipete:
Maintainer review complete; I added the changelog entry in a384ecc without changing the contributor implementation.

Disposition: LAND.

Why: current main performs a single exact chat/message lookup, while authenticated startup can migrate direct-message history between phone-number and LID chat identities. This branch retries only after `sql.ErrNoRows`, only for direct PN/LID counterparts, and passes the resolved alias through the self-authored participant calculation before building the outgoing quote. Plain sends, group replies, and non-missing database errors keep their existing behavior.

Proof:

- `go test ./cmd/wacli -run 'Test(BuildTextReplyContextInfoFindsQuoteUnderChatAlias|SendTextReplyToOwnMessageUnderChatAliasUsesLIDParticipant)' -count=1 -v` — both alias directions and the self-authored LID participant regression pass.
- Full gate green: `pnpm format:check`, `pnpm lint`, plain Go tests, FTS5 tests, Windows lock cross-compile, CGO guard, docs/release tests, `pnpm build`, `pnpm docs:site`, and `git diff --check`.
- Built CLI: `./dist/wacli --version` prints `wacli 0.15.0`; `send text --help` exposes `--reply-to` and `--reply-to-sender`.
- Source-blind behavior probes confirmed structured failures for missing input and an unauthenticated fresh store, with no false success. Authenticated sending was not repeated because that would mutate a personal account; the PR body contains the contributor's redacted disposable-account delivery proof.
- Codex branch autoreview: clean, no accepted/actionable findings (overall correctness 0.98). Changelog-only pre-commit review also clean (0.99).

Trust context: @0xlucuma is a new GitHub account and this is their first wacli PR, so I reviewed the full send/identity path rather than relying on trust. The implementation and proof held up.

@clawsweeper re-review

PR #328, opened 2026-08-02
fix(send): surface local store failures after successful file send

SebTardif:
## What Problem This Solves

After a successful WhatsApp file send, `send_file` discarded `UpsertChat` / `UpsertMessage` / `UpsertStatusMessage` errors (`_ = ...`). Disk full, locked DB, or corruption left local history silently wrong while the network send succeeded.

### After

```go
if err := a.DB().UpsertMessage(...); err != nil {
  return id, nil, fmt.Errorf("message sent but local message store update failed: %w", err)
}
```

```console
$ go test ./cmd/wacli/ -count=1
ok  github.com/openclaw/wacli/cmd/wacli
$ go build ./...
```

## Real behavior proof

- **Behavior or issue addressed:** Surface local SQLite persist failures after a successful media send instead of swallowing them.
- **Real environment tested:** macOS arm64, Go, `/tmp/oc-batch-20260802/wacli`.
- **Exact steps or command run after this patch:** `go test ./cmd/wacli/ -count=1` and `go build ./...`.
- **Evidence after fix:** Package tests pass; build clean.
- **Observed result after fix:** Store errors return with message id still available when send completed.
- **What was not tested:** Live WhatsApp send against a forced SQLite failure (would need a broken DB path in integration).

## Summary

Check Upsert* errors after SendProtoMessage; wrap with clear "sent but store failed" errors.

### DIFF
--- a/cmd/wacli/send_file.go
+++ b/cmd/wacli/send_file.go
@@ -159,8 +159,10 @@ func sendFile(ctx context.Context, a interface {
 		return "", nil, err
 	}
 
+	// Network send already succeeded. Local store failures must still surface so
+	// history does not silently diverge (disk full, locked DB, corruption).
 	if to == types.StatusBroadcastJID {
-		_ = a.DB().UpsertStatusMessage(store.UpsertStatusMessageParams{
+		if err := a.DB().UpsertStatusMessage(store.UpsertStatusMessageParams{
 			MsgID:         id,
 			Timestamp:     now,
 			FromMe:        true,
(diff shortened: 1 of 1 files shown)


steipete:
Thanks @SebTardif — the defect you found was real: after a successful media send, all local history upserts were silently discarded, so disk-full or locked-store failures left local history quietly diverged.

Review surfaced one design issue with returning a hard error: every `sendFile` caller (file, voice, status, and the IPC delegation path) drops the returned message id on error, so a post-delivery store failure would report a *delivered* message as a total failure — and retrying automation would double-send it.

Landed as #333 (merge fd7179a530bb, with co-author credit): `sendFile` now returns the delivered id together with a `storeWarning`; every boundary keeps exit code 0, prints a stderr warning, and exposes `store_warning` in JSON/IPC output. Regression tests force a real closed-SQLite failure with a stub client delivering the message. The same swallowed-upsert pattern in sticker/reaction/poll/text sends is queued as a follow-up.

Closing in favor of #333.

PR #338, opened 2026-08-03
feat: add WhatsApp locations (send/parse/store)

0xlucuma:
## What

Adds location support in both directions:

- `wacli send location` sends a native WhatsApp pin.
- Inbound `LocationMessage` and `LiveLocationMessage` are parsed instead of discarded, and their coordinates are stored in a new `message_locations` table (migration 25).

## Why

`extractMedia` had no branch for `LocationMessage`, so a received pin reached the store with no media type, no text and no coordinates. `baseDisplayText` then fell through to its empty-string case and the row was written as `(message)` — the same placeholder used for undecryptable content. The coordinates were not stored anywhere, so nothing downstream could recover them.

`mediaLabel` already had a `case "location": return "location"` that nothing could reach, which suggests the media type was intended and only the parser was missing.

There was also no way to send a pin.

## What changed

- `internal/wa/messages_media.go` — `extractLocation` parses `LocationMessage` (name, address) and `LiveLocationMessage` (caption, `is_live`), setting a `location` / `live_location` media type plus structured coordinates on `ParsedMessage`. No `direct_path` or `media_key` is set, so a pin never enters the media download queue.
- `internal/app/sync.go` — persists the parsed location alongside the message row; `mediaLabel` gains a `live_location` case so the shared display path reads "Sent live location" rather than the wire name.
- `internal/store/` — new `message_locations` table keyed `(chat_jid, msg_id)`, following the `polls` table rather than widening the `messages` upsert. Registered in `ensureCurrentSchema` so an interrupted migration self-repairs, and dropped with its chat in `DeleteChat` / group deletion alongside `polls`/`starred`.
- `cmd/wacli/send_location_cmd.go` — the new verb, plus a `location` delegate kind in `send_ipc.go` so it works while `sync --follow` holds the store lock.
- `docs/send.md`, `docs/spec.md`, `docs/store.md`, `docs/sync.md`.

`internal/store/storedb/*` is generated: the hand-written change is in `internal/store/sqlc/queries.sql`, regenerated with `pnpm generate:sqlc`.

## New command

```bash
wacli send location --to RECIPIENT --latitude LAT --longitude LNG [--name TEXT] [--pick N] [--post-send-wait 2s]
```

```console
$ wacli --json send location --to REDACTED --latitude 51.4779 --longitude -0.0015 --name "Royal Observatory"
{"success":true,"data":{"id":"3EB0FEE3…","sent":true,"to":"REDACTED@lid"},"error":null}
```

Both coordinate flags are required and must be set explicitly: `0,0` is a real coordinate in the Gulf of Guinea, so a zero default cannot be distinguished from an omitted flag. NaN and infinities are rejected, latitude is bounded to ±90 and longitude to ±180. Validation runs before the store is opened.

`LocationMessage` carries no caption, mentions or reply context, so the command deliberately exposes none of those flags.

## Storage

Migration 25 creates:

```sql
CREATE TABLE IF NOT EXISTS message_locations (
    chat_jid  TEXT NOT NULL,
    msg_id    TEXT NOT NULL,
    latitude  REAL NOT NULL,
    longitude REAL NOT NULL,
    name      TEXT,
    address   TEXT,
    is_live   INTEGER NOT NULL DEFAULT 0,
    PRIMARY KEY (chat_jid, msg_id)
);
```

No existing table is altered and the `messages` upsert is untouched. Verified on a copy of a real 5.6k-message store: schema version 24 → 25, message count unchanged, `PRAGMA integrity_check` = `ok`.

## Tests

- `internal/wa/messages_location_test.go` — static pin, live share, non-location messages, and the top-level parse path that used to drop coordinates.
- `internal/store/locations_test.go` — upsert/get round trip, live-location updates replacing position, out-of-range rejection, delete, and that deleting a chat drops its locations while leaving another chat's intact.
- `internal/store/schema_test.go` — the table and its columns exist after migration, and a store whose `message_locations` was dropped while the version row remained is repaired on open (mirrors `TestOpenRepairsRecordedCallEventsMigrationMissingTable`).
- `cmd/wacli/send_location_cmd_test.go` — flag validation table, message construction, recipient plumbing, and outbound persistence canonicalising a LID chat to its phone JID.
- `cmd/wacli/send_store_warning_test.go` — location persistence follows the `store_warning` partial-success contract from #333/#335.

Full gate, green on go1.26.5:

```
```

Also ran `pnpm docs:site` and the `deadcode` and `govulncheck:source` steps from `ci.yml`.

## Real behavior proof (live account, redacted)

Binary built from this branch, paired to a real account whose `sync --follow` was running throughout — so the send below took the IPC delegate path, not the direct-store path. Note the resolved recipient comes back as a `@lid`, which is why the outbound persist canonicalises it back to the phone JID.

A pin was sent with the command above, and one was sent **from the phone** to that account. Contact name, JIDs, store path and message ids are redacted or truncated; the inbound pin's coordinates are the sender's real position and are redacted too.

```console
$ wacli --read-only --store REDACTED messages list --chat REDACTED --limit 5
TIME                 CHAT      FROM  ID                      TEXT
2026-08-03 15:47:56  REDACTED  me    3EB0FEE3…  Sent location
2026-08-03 12:51:27  REDACTED  me    3EB0C0F3…  Sent location
2026-08-03 12:50:41  REDACTED  THEM  3A33C6…    Sent location
2026-08-03 11:51:18  REDACTED  THEM  3A74D5…    (message)
```

The contrast is the point. `11:51:18` is an inbound pin received by the **pre-change** binary: no coordinates, rendered as `(message)`. `12:50:41` is an inbound pin received by this branch.

Stored rows for those messages:

```console
sqlite> SELECT m.from_me, m.media_type, m.display_text, l.latitude, l.longitude, l.name, l.is_live
   ...> FROM messages m JOIN message_locations l
   ...>   ON l.chat_jid = m.chat_jid AND l.msg_id = m.msg_id ORDER BY m.ts;
0|location|Sent location|REDACTED|REDACTED||0                    <- received from the phone
1|location|Sent location|51.4779|-0.0015|Royal Observatory|0     <- sent by the command above

sqlite> SELECT version, name FROM schema_migrations ORDER BY version DESC LIMIT 1;
25|message locations
```

The inbound row carries an empty `name` because the sender attached no place label — WhatsApp sends only coordinates in that case.

Pins received before migration 25 cannot be backfilled: the protobuf is not retained, so `3A74D5…` above stays `(message)` permanently.

## Notes

- Replying to a location is still rejected with `unsupported stored media type "location"`. That is pre-existing behaviour and is already asserted by `TestSendTextMessageRejectsUnconstructableQuotesBeforeSending`, which seeds exactly that media type. Making pins quotable would need a `LocationMessage` quote builder and is out of scope here.
- Live locations are stored as their latest reported position; subsequent updates overwrite the row. Whether whatsmeow delivers a stream of updates for a live share was not exercised against a live account.


0xlucuma:
Fixed in `81c912b` — thanks, the purge gap was real.

`PurgeMessage` now deletes `message_locations` in the same transaction as `polls`, `poll_votes`, `starred` and `message_local_media_aliases`, so a purged message no longer leaves its coordinates behind. `TestPurgeMessageClearsStoredLocation` covers it; reverting the one-line delete makes it fail with `purged message retained its coordinates`.

On the foreign key: I left it off deliberately. Four of the five per-message side tables (`polls`, `poll_votes`, `starred`, `message_payload_purges`) have no FK to `messages` and rely on the explicit deletes in `PurgeMessage`; only `message_local_media_aliases` declares one. `message_locations` follows the `polls` shape it was modelled on. Happy to add the cascade instead if you would rather converge on the alias table's pattern.

Note that purge nulls the message payload rather than deleting the row, so a cascade would not have covered this case on its own — the explicit delete is required either way.

@clawsweeper re-review

0xlucuma:
Fixed in `84b1742` — this one was a good catch, and it is the more dangerous of the two.

`MigrateLIDToPN` moved `messages`, `message_payload_purges`, `message_local_media_aliases` and `polls` to the phone identity and deleted the LID rows, but nothing moved `message_locations`. After a LID-to-phone migration the coordinates stayed keyed to the dead LID chat, which means they were both unreachable (lookups use the new chat JID) and unreachable by cleanup (`PurgeMessage` and `DeleteChat` key on the new JID too) — so the purge fix from the last round would not have covered them either.

`migrateLIDMessageLocationsToPN` now mirrors `migrateLIDPollsToPN`: copy to the phone JID, skip rows whose message is in the purge ledger under either identity, delete the LID rows, then drop any destination row whose message was already purged.

Two regression tests:

- `TestMigrateLIDToPNMovesMessageLocations` — coordinates and name arrive under the phone JID and the LID row is gone. Fails without the fix with `GetMessageLocation pn: sql: no rows in result set`.
- `TestMigrateLIDToPNDropsPurgedMessageLocations` — a location purged under the LID identity does not resurface under the phone identity.

Full gate green.

0xlucuma:
One more in `68e4f8b`, found by auditing rather than waiting for the next review.

After the LID finding I enumerated every site that handles a per-message side table, using `polls` as the reference since `message_locations` was modelled on it. That turned up a fourth path I had missed: `DeleteChatsOlderThan` (`store cleanup --days N`) deletes `poll_votes`, `polls` and `starred` for stale chats, but not locations — so a cleanup would have left coordinates behind for chats it had just removed.

`TestDeleteChatsOlderThanDropsMessageLocations` covers it, and asserts a recent chat keeps its pin. Fails without the fix with `stale chat retained its coordinates`.

For the record, the five paths and their current state:

| path | locations handled |
|---|---|
| `PurgeMessage` | yes (`81c912b`) |
| `MigrateLIDToPN` | yes (`84b1742`) |
| `DeleteChatsOlderThan` | yes (`68e4f8b`) |
| `DeleteChat` | yes (in the original commit) |
| `DeleteGroupLocalData` | yes (in the original commit) |

Full gate green.

@clawsweeper re-review

0xlucuma:
Fixed in `fd4ba2c`, and you were right that it was outstanding from the prior cycle — I read the findings list too narrowly and missed it twice. Sorry for the round trip.

Parsing `LiveLocationMessage` without teaching the two helpers about it left live shares half-supported: `contextInfoForMessage` had a `GetLocationMessage` branch but no live one, so a reply to a live share dropped its `StanzaID` and `Participant`, and `displayTextForProto` fell through to `""`, so quoting one rendered a blank label.

Both helpers now have a `GetLiveLocationMessage` branch, returning the message's `ContextInfo` and the label `Sent live location` respectively.

Three regression tests, each failing on the pre-fix head:

- `TestParseWAProtoKeepsLiveLocationReplyContext` — reply id and sender survive (`reply id = ""` before).
- `TestDisplayTextForProtoLabelsLiveLocation` — `display text = ""` before.
- `TestQuotedLiveLocationRendersALabel` — end to end: replying to a live share yields `Sent live location` as the quoted preview rather than an empty string.

On the remaining items: the lifecycle risk is now covered across all five paths (`PurgeMessage`, `MigrateLIDToPN`, `DeleteChatsOlderThan`, `DeleteChat`, `DeleteGroupLocalData`), each with a test that fails without its fix. The product-scope question is a maintainer call and I have nothing to add to it beyond the PR body.

@clawsweeper re-review

0xlucuma:
On the retention question — my read is that a pin is message content, not a separate payload class, so it should persist like the rest of a message. Reasoning, for whatever it is worth to the decision:

WhatsApp delivers a location as a message the sender chose to send, and wacli already retains every message body, contact name and phone number, status broadcast and call record by default. Gating coordinates while `messages.text` stores verbatim would be an odd line to draw.

Media bytes are the one existing opt-in, but I do not think they are the right analogue: `--download-media` exists for size and bandwidth, which is why it has a `media backfill` escape hatch. A location has no equivalent recovery — the protobuf is not retained, so anything skipped is unrecoverable, and a forgotten flag would silently and permanently lose data that the message itself still claims to contain.

The counter-argument I would accept is that coordinates are more directly identifying than most text, and that is a values call rather than a technical one.

What I have added in `79030fe` is documentation rather than behaviour: the retention was previously stated without the corresponding deletion story. `docs/store.md` and `docs/send.md` now say that coordinates follow the same lifecycle as any other retained payload — removed by `messages purge`, removed by `store cleanup` and `chats cleanup`, and moved rather than stranded by a LID-to-phone migration. That is the property that makes storing them by default defensible, and it was implemented but undocumented.

Happy to add an opt-in flag instead if you would rather draw the line differently; it is a small change and I did not want to presume the policy.

Issue #343, opened 2026-08-06
Incoming message edits stored as an empty `(message)` row instead of updating the original (live path, v0.15.2)


dvainrub:
## Summary

Incoming message edits are stored as a **new, content-less row** (`text = NULL`, `display_text = "(message)"`) instead of updating the original message. The original row keeps its **pre-edit text**.

This is closely related to #245, which was scoped to history/backfill and closed by 72db4db. The repro below is on the **live path** (`sync --follow` connected, message received in real time) with **v0.15.2**, so the live path is affected too.

## Why this is worse than it looks

The failure is silent and produces a *plausible* wrong answer rather than a visible gap. Any consumer reading local history gets the **superseded text**, with no indication that it was replaced. A missing message is noticeable; a stale one that reads perfectly is not.

Concretely: an automated summarizer over this database reported that a contact "wants to integrate Lightspeed", quoting the stored text verbatim. The contact had edited that message to say **Instagram** seconds after sending it. Nothing in the JSON output flagged the discrepancy, and the stale value happened to be plausible in context, so it was not questioned.

## Environment

- `wacli` **0.15.2** (Homebrew, `openclaw/tap`), macOS 26.5.1 arm64
- `go.mau.fi/whatsmeow v0.0.0-20260722203353-e9a033b24933`
- `wacli sync --follow` running as a launchd agent, connected at the time of the edit
- Freshly upgraded from 0.13.0 and daemon restarted on the new binary before the repro

## Steps to reproduce

1. Run `wacli sync --follow` and confirm it is connected.
2. From another WhatsApp account, send a 1:1 text message (`Hola`).
3. Edit that message in the WhatsApp app a few seconds later.
4. Inspect the store.

## Observed

```sql
SELECT msg_id, datetime(ts,'unixepoch') ts, from_me, edited, edited_ts,
       quoted_msg_id, text, display_text
FROM messages WHERE chat_jid = '<redacted>@s.whatsapp.net' ORDER BY ts DESC LIMIT 2;
```

```text
        msg_id = A546F3BA8D8D41E06FDC98B43EC77149   <-- the edit
            ts = 2026-08-06 04:30:10
       from_me = 0
        edited = 0
     edited_ts = 0
 quoted_msg_id =
          text = NULL
  display_text = (message)

        msg_id = A5186118AA9167B8CAFE2CAC1770964D   <-- the original, NOT updated
            ts = 2026-08-06 04:30:04
       from_me = 0
        edited = 0
[...]

PR #344, opened 2026-08-06
sync: name the payload when a message is stored without content

dvainrub:
Makes content-less message rows diagnosable. Opening as a PR because #343 was closed automatically and I could not reopen it; the full context lives below so it does not depend on that thread.

## The gap

When parsing extracts nothing from a message, it is stored with a `(message)` placeholder. That row is indistinguishable from a message that genuinely carried no content, so there is no signal that anything was dropped and no way to tell what it was.

On one account: **6,403 of 64,349 stored messages (10%)**, and **3.11% over the last 14 days**. None of them join to `polls` or `call_events`, so the known content-less types do not account for them.

## Why it is worse than a gap

At least some of these are unhandled message edits. When an edit is not unwrapped, the original row keeps its **pre-edit text**, which reads as perfectly valid. A missing message is noticeable; a superseded one is not.

Concretely, an automated summariser over this database reported that a contact "wants to integrate Lightspeed", quoting the stored text verbatim. The contact had edited that message to say **Instagram** seconds after sending it. Nothing in the output flagged a discrepancy.

## The behaviour this came from (context from #343)

Reproduced on **v0.15.2** with `sync --follow` connected: an incoming 1:1 text (`Hola` at `04:30:04Z`), edited at `04:30:10Z`, produced a second row with its own `msg_id`, `text = NULL`, `display_text = "(message)"`, `edited = 0`, while the original kept `Hola`.

Two details from that investigation, in case they help:

- **It was the live path, not history.** The daemon connected at `04:20:31Z` and `sync.err` ends at `Connected.` with no `Synced N messages...` line after it, so no catch-up batch ran between connection and the message ten minutes later.
- **The handling code is present and unchanged.** `extractProtocolMutation` has handled `MESSAGE_EDIT` since 72db4db (May), and it is byte-for-byte identical between `v0.15.2` and current `main` — the only change to `extractWAProto` in between is the added `extractLocation(m, pm)` line from #338. So the code exists; something upstream of it is not reaching it. **This PR does not attempt to fix that**, because I have not identified the cause and would rather not guess at it.

`edited = 1` rows are also worth a look: 186/month in April and May 2026, 47 in June, 1 in July, 0 in August, while placeholder rows keep being produced throughout.

## What this PR does

Records which `waE2E.Message` field was populated when extraction produced nothing, and emits a warning at store time with the chat, message id and field name. **Nothing about what gets stored changes.**

Field names come from the protobuf descriptor, so a new WhatsApp message type is reported without needing a code change here.

Deliberately **not** flagged, to keep the warning from becoming noise that gets tuned out:

- messages carrying only `messageContextInfo`
- empty messages
- revokes, which are content-less by design

## Tests

Six cases in `internal/wa/unhandled_payload_test.go`: both parse entry points (live and history), a handled payload staying unflagged, and one per exclusion above.

## Gate

`gofmt -l` clean · `go vet ./...` · `go test ./...` · `go test -tags sqlite_fts5 ./...` · `go build -tags sqlite_fts5 ./cmd/wacli` · `git diff --check` — all pass on this branch.

## Unrelated, but blocking verification for tap users

`openclaw/homebrew-tap` is still pinned to `wacli 0.15.2`: the `update-formula` workflow has no run after 2026-08-03, so it never fired for v0.16.0 despite that release being complete and public. Filed as openclaw/homebrew-tap#13. It means anyone on the tap cannot install v0.16.0 to check the behaviour above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- Sync: identify unhandled payload types when content extraction produces a placeholder, with bounded diagnostics during history replay. (#344 - thanks @dvainrub)
+
 ## 0.16.0 - 2026-08-02
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- Sync: warn when a message is stored without content, naming the payload field that went unextracted, so the resulting `(message)` placeholder can be diagnosed instead of silently losing the body. (#343)
+
 ## 0.16.0 - 2026-08-02
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,5 @@
 # Changelog
 
-## Unreleased
-
-### Added
-
-- Sync: warn when a message is stored without content, naming the payload field that went unextracted, so the resulting `(message)` placeholder can be diagnosed instead of silently losing the body. (#343)
-
 ## 0.16.0 - 2026-08-02
 
 ### Added
--- a/internal/app/sync.go
+++ b/internal/app/sync.go
@@ -399,6 +399,7 @@ func chatKind(chat types.JID) string {
 
 func (a *App) storeParsedMessage(ctx context.Context, pm wa.ParsedMessage) error {
 	pm.Chat = a.canonicalStoreJID(ctx, pm.Chat)
+	a.warnUnhandledPayload(pm)
 	chatJID := canonicalJIDString(pm.Chat)
 	chatName := a.wa.ResolveChatName(ctx, pm.Chat, pm.PushName)
 	if pm.Chat != types.StatusBroadcastJID {
@@ -712,6 +713,28 @@ func (a *App) buildDisplayText(ctx context.Context, pm wa.ParsedMessage) string 
 		base = "(message)"
 	}
 	return base
+}
+
+// warnUnhandledPayload surfaces messages whose payload produced no content, so
+// the resulting "(message)" placeholder is diagnosable. Without it the row is
+// indistinguishable from a message that genuinely carried nothing, and any
+// consumer reading local history silently sees a gap it cannot account for.
+func (a *App) warnUnhandledPayload(pm wa.ParsedMessage) {
+	if pm.UnhandledPayload == "" {
+		return
+	}
+	a.emitWarning(
+		"unhandled_message_payload",
+		fmt.Sprintf(
+			"stored message %s in %s without content: unhandled payload %s",
+			pm.ID, canonicalJIDString(pm.Chat), pm.UnhandledPayload,
+		),
+		map[string]any{
+			"chat_jid": canonicalJIDString(pm.Chat),
+			"msg_id":   pm.ID,
+			"payload":  pm.UnhandledPayload,
+		},
+	)
 }
 
 func baseDisplayText(pm wa.ParsedMessage) string {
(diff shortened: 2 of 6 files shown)


dvainrub:
Thanks — the two P2s were real defects, both fixed in 2392ba2.

**Nested wrappers.** Correct, and it broke the diagnostic precisely where it was needed: an unhandled edit would report `protocolMessage` instead of the field that went missing. `extractWAProto` now returns the leaf it extracted from and the callers name that, rather than re-walking the wrappers in a second pass — deriving the leaf from the same traversal keeps the two from drifting apart later.

The three new tests are red on the previous behaviour and green after, with exactly the wrapper names predicted:

```
--- FAIL: TestMarkUnhandledPayloadNamesLeafInsideDeviceSent
    expected the leaf payload, got "deviceSentMessage"
--- FAIL: TestMarkUnhandledPayloadNamesLeafInsideProtocolEdit
    expected the leaf payload, got "protocolMessage"
--- FAIL: TestMarkUnhandledPayloadNamesLeafInsideEditedMessage
    expected the leaf payload, got "editedMessage"
```

A fourth covers a wrapper around understood content staying unflagged, so the fix cannot be satisfied by flagging everything.

**Warning timing.** Also correct. It now runs after `UpsertMessage` succeeds rather than before the chat and message upserts, so a failed write no longer produces a success-shaped diagnostic.

**Changelog.** Dropped.

**On the runtime proof.** Understood, and I would rather not fake it. Two honest constraints:

1. The warning fires on ingestion, so producing it requires a live WhatsApp session on a real account rather than a fixture. I can run the patched binary against my own account and post the redacted stderr line — say so and I will.
2. Separately, anyone verifying via Homebrew currently cannot: the tap is still pinned to `wacli 0.15.2` because `update-formula` never fired for v0.16.0 (openclaw/homebrew-tap#13).

Gate re-run on the new head: `gofmt -l` clean, `go vet ./...`, `go test ./...`, `go test -tags sqlite_fts5 ./...`, `go build -tags sqlite_fts5 ./cmd/wacli`, `git diff --check`.

steipete:
Maintainer repair pushed as 137a8ba without rewriting the contributor commits.

The history-sync warning path now emits one detailed example per payload shape, caps detailed examples at ten per history pass, and emits one aggregate suppression summary. Live ingestion keeps the specific post-store diagnostic.

Proof:

- focused `go test ./internal/wa ./internal/app` passed
- full repository gate passed: format, vet, ordinary + FTS tests, Windows lock cross-compile, CGO guard, docs tests, build, and diff check
- real built `wacli 0.16.0` connected to an authenticated account, synced 25 messages, and exited successfully without sending; the production path emitted seven redacted `unhandled_message_payload` warnings naming two payload shapes
- source-blind behavior validation passed against the built CLI and redacted live evidence
- Codex autoreview of both the maintainer delta and complete PR branch was clean with no accepted/actionable findings

steipete:
Landed as f5f5c6f6846dc702ab99d026d40f2c54f9dffc2e with contributor credit preserved.

Final verification:

- focused parser/app tests passed
- full local gate passed: formatting, vet, ordinary + FTS suites, Windows lock cross-compile, CGO guard, docs tests, build, and diff check
- built `wacli 0.16.0` connected to an authenticated account, synced 25 messages, and exited cleanly without sending; seven redacted production warnings named two actual unhandled payload shapes
- source-blind behavior validation passed
- exact-head CI for 137a8ba23665f71cd8ac7e2fcc4a1e57f37a4523 passed: test, Linux release builds, and Docker
- complete-branch Codex autoreview was clean with no accepted/actionable findings

History replay now reports one example per payload shape, caps detailed examples at ten per pass, and emits one suppression summary rather than flooding operator output.

PR #363, opened 2026-08-15
fix(sync): apply authenticated encrypted message edits

### LINKED ISSUE #362: SecretEncryptedMessage{MESSAGE_EDIT} edits are stored as opaque rows instead of being applied
### Summary

WACLI does not appear to process WhatsApp message edits when the edit arrives wrapped as `SecretEncryptedMessage{MESSAGE_EDIT}`.

In a controlled live test, the original message remained unchanged and unedited, while the edit was stored as a separate opaque `secretEncryptedMessage` row. That row contained no usable message body, edit flag, original-message relationship, or revision identity exposed through WACLI's stored output.

This prevents downstream consumers from reconstructing the corrected message without inventing data.

### Environment

- WACLI: `v0.17.0` (pinned release)
- Platform: macOS
- WhatsApp account: LID-migrated account
- Capture mode: `wacli sync` against an allowlisted test chat

A review of the `v0.17.1` source found calls to `DecryptSecretEncryptedMessage` for encrypted poll options, but no equivalent handling for encrypted message edits.

### Minimal reproduction

1. Start WACLI sync for a chat using a WhatsApp account that emits encrypted edit envelopes.
2. Send an ordinary text message from another participant.
3. Edit that message in WhatsApp.
4. Inspect WACLI's stored message records/output.

### Expected behaviour

WACLI should decrypt the `SecretEncryptedMessage` payload, pass the resulting `ProtocolMessage{MESSAGE_EDIT}` through its existing edit-processing path, and retain enough lineage to identify the original message and the edited revision.

### Actual behaviour

- The original message remains unchanged.
- Its edit state remains false/unset.
- A separate `secretEncryptedMessage` record is stored.
- The stored edit record has no usable corrected body or original-message linkage.
- Restarting sync and replaying the captured history does not reconcile the edit.

### Related work

- WACLI issue #245 and PR #246 handled ordinary `ProtocolMessage{MESSAGE_EDIT}` edits during history/backfill, but do not appear to cover the newer encrypted wrapper:
  - https://github.com/openclaw/wacli/issues/245
  - https://github.com/openclaw/wacli/pull/246
- A related whatsmeow-based project implemented handling for LID-migrated accounts by decrypting `SecretEncryptedMessage{MESSAGE_EDIT}` and then passing the result through its existing edit path:
  - https://github.com/aldinokemal/go-whatsapp-web-multidevice/pull/693

### Suggested direction

When `SecretEncType == MESSAGE_EDIT`:

1. call the existing `DecryptSecretEncryptedMessage` implementation;
2. validate that the decrypted payload contains `ProtocolMessage{MESSAGE_EDIT}`;
3. route it through the same edit reconciliation path used for ordinary protocol-message edits; and
4. add regression coverage for live sync and history/backfill, including original-message linkage and edited-body persistence.

The original reporter can retest a pinned release or source revision against the same controlled reproduction once a candidate fix is available.

goutamadwant:
Encrypted WhatsApp edits were parsed as opaque `secretEncryptedMessage` rows, leaving the original text unchanged. Live sync and history/backfill now decrypt `MESSAGE_EDIT` envelopes before the existing edit reconciliation path, updating the original row and preserving its timestamp and sender. Fixes #362.

The update is bound to the envelope's authenticated sender/chat and target ID. The SDK's secret lookup requires the matching original sender (including LID/phone aliases); successful decryption is followed by an explicit editor/owner check. Sender-relative `FromMe` is normalized to the local account's direction. Redirected targets, unsupported payloads, and nested protocol mutations produce warnings without storage changes or edit webhooks. The implementation and tests are isolated in `sync_edits*.go`.

Validation: full repository gate (format, vet, both dead-code modes, plain/FTS tests, Windows cross-build, CGO-required check, docs/release-script tests, executable build, documentation site); focused race tests.

Compiled-CLI proof used synthetic WhatsApp transport with the pinned SDK's real SQL secret lookup, LID mapping, HKDF/AES-GCM decryption, history envelope parser, production SQLite persistence, and a local HTTP webhook receiver. No real WhatsApp account was accessed. The baseline was main `dfef73cc0f724b5eefe4946417f52496ce466cca`.

| Scenario | Baseline | Fixed |
| --- | --- | --- |
| Group edit, live + newest-first history | Original unchanged; separate opaque envelope row | One original row, corrected body, edited flag, original timestamp |
| Incoming DM, outgoing DM, sender-relative `FromMe`, LID/phone alias | — | All accepted in live/history; sender and direction preserved |
| Linked-device and ephemeral wrappers | — | Decrypted after SDK unwrapping, including newest-first history |
| Different author with a valid shared secret | — | SDK decrypts; application rejects; original untouched |
| Forged `FromMe` or participant | — | SDK sender-scoped lookup rejects |
| Redirected chat, nested revoke, target-ID mismatch, unsupported or empty payload | — | Rejected; original untouched |
| Missing secret or modified ciphertext | — | Rejected; original untouched |

Thirty-eight executable runs passed: two baseline reproductions and 36 fixed scenarios. All preserve unrelated history. Accepted live edits produce one edit webhook carrying the original ID and corrected body; rejected edits produce none. Histories remain webhook-free. Unit regressions also exercise the real SDK for 34 live/history combinations. Synthetic proof does not establish live transport delivery or recovery of already-opaque historical rows; re-fetch depends on what WhatsApp supplies.

Build/run commands: `go build -overlay <synthetic-transport-overlay.json> -tags sqlite_fts5 ./cmd/wacli`; `wacli --store <synthetic-store> --json sync --once --idle-exit 50ms --webhook <local-receiver> --webhook-allow-private`. Observed baseline: `rows=2, original_text="original body", edited=0`; fixed: `rows=1, original_text="edited body", edited=1`. Rejected variants: `rows=1, original_text="original body", edited=0`, no edit webhook.

Thanks @goutamadwant for the implementation, @dvainrub for identifying the authorization boundary, and @Entretoize for the earlier live wire-format evidence in this PR. No CLI flags, schema, or version changes.

Disclosure: AI was used to understand the codebase and review the fix.


Entretoize:
Validated the premise of this fix against a **live WhatsApp account**, which the PR description lists as a gap ("No live WhatsApp account was used for validation").

## Method

I did not build the branch. Instead I reproduced the issue's steps against a real linked device and instrumented the wire format directly, using a small standalone whatsmeow harness sharing the same session store — pinned to the exact whatsmeow revision wacli depends on (`v0.0.0-20260806224404-e277b766ab33`). It sends nothing; it only listens and decodes.

Steps: send a message from the primary device, wait, edit it. Environment: wacli 0.17.1 store, Windows, LID-migrated account (same as #362).

## Result

```
[10] 10:00:51  Message chat=<redacted>@lid  id=AC7CCA3D3F222666380A17996BFC88C3
[12] 10:01:13  SecretEncryptedMessage{MESSAGE_EDIT}
       envelope id     AC172097740B348B9B95BC9D971B1B9C
       target key      AC7CCA3D3F222666380A17996BFC88C3
       decrypt         OK
       edited text     "<edited body recovered verbatim>"
       target match    consistent (payload key == envelope key)
```

Each assumption this patch relies on holds on real traffic:

1. the edit **does** arrive wrapped as `SecretEncryptedMessage{MESSAGE_EDIT}` (not the plain `ProtocolMessage` form)
2. `DecryptSecretEncryptedMessage` **succeeds** on it
3. the decrypted payload **is** a `ProtocolMessage{MESSAGE_EDIT}` carrying the edited body
4. envelope `TargetMessageKey` and decrypted `ProtocolMessage.Key` **agree** — so the mismatch branch is a genuine guard, not the normal path

Point 4 is worth noting: `protocol.Key = target` in `decryptSecretEdit` is a no-op on healthy traffic here, which is the behaviour you'd want.

## Why the opaque rows are unrecoverable after the fact

The envelope carries its **own** message ID, distinct from the target's. Without decryption there is no stored link between the two, and wacli currently persists neither the decrypted body nor the target key — so the corrected text is lost at parse time, not merely unrendered.

On this store that amounts to **438 contentless rows** in a single 29,634-message chat (they also include unparsed albums, so the edit share is a subset). I tried to recover them by re-requesting history: the on-demand bundles for those ranges come back empty, so a fix cannot be retroactive for edits already synced. That makes the loss permanent for existing installs and, in my view, raises the value of merging this.

## Scope of what I verified

- ✅ wire format, decryptability, payload shape, target-key consistency — on live traffic
- ❌ wacli's persistence path (that the original row is updated in place, `edited` set, no extra row) — not verified; that still rests on this PR's fixtures. Building wacli needs a CGO toolchain I do not have on this machine.

Happy to re-run against a built binary if a prebuilt artifact of this branch is available anywhere.

Entretoize:
**Correction to my previous comment.** I claimed the loss is permanent for already-synced edits:

> I tried to recover them by re-requesting history: the on-demand bundles for those ranges come back empty, so a fix cannot be retroactive for edits already synced.

That is wrong, and I apologise for putting an unverified claim in this thread. I had generalised from on-demand requests that returned nothing for an unrelated reason (unresolvable anchor keys — see #371), never testing the case properly.

## What the controlled test actually shows

I deleted a chat's local rows and re-fetched them through `history backfill`, then compared.

**Edits that wacli already applied come back correctly.** Three rows with `edited=1` were deleted and re-fetched: all three returned with the *corrected* body and the flag intact. So history sync serves the phone's **current** message state, not the pre-edit original.

**Contentless rows come back contentless — but the payload is re-delivered.** Four `(message)` rows were deleted and re-fetched. They returned still empty, and the sync log shows why:

```
stored message AC05BF6B… without content: unhandled payload albumMessage
```

The bundle carries the payload. wacli receives it, cannot decode it, and drops it — deterministically, the same way each time. Nothing is missing on the wire; the gap is in the parser.

## Why this matters for this PR

A fix here **is** retroactive. Once merged, an existing install can delete the affected range and re-run `history backfill`, and the previously opaque edits will be decoded and applied on the way in. That is a stronger argument for merging than the one I gave, not a weaker one — my earlier framing understated the value.

It also means the same holds for the payload types tracked in #365 (`albumMessage` and friends): those bodies are still being served on every re-fetch, and are recoverable the moment there is a decoder for them.

Everything in my previous comment about the wire format — envelope shape, successful decryption, payload type, target-key consistency on live traffic — stands unchanged. Only the retroactivity claim was wrong.

dvainrub:
I arrived at the same diagnosis independently and opened #407 before noticing this PR, which predates mine by a month and has the wider scope (history sync, target validation, per-failure warnings). I'm closing mine in favour of this one.

Handing over the one thing my investigation adds that isn't here: **decrypting an edit does not establish that the editor owns the message being edited.** I believe this branch is exposed to it too.

## Why successful decryption is not authorization

The message secret is shared with recipients, and whatsmeow derives the edit key from the envelope sender and the original sender as **separate** inputs, without requiring them to match:

```go
// msgsecret.go, decryptMsgSecret
origSender, err := getOrigSenderFromKey(msg, origMsgKey)
baseEncKey, storedOrigSender, err := cli.Store.MsgSecrets.GetMessageSecret(ctx, msg.Info.Chat, origSender, origMsgKey.GetID())
secretKey, additionalData := generateMsgSecretKey(useCase, msg.Info.Sender, origMsgKey.GetID(), origSender, baseEncKey)
```

So in a group, any participant holding `baseEncKey` for someone else's message can construct an envelope that decrypts cleanly against it.

## Why the checks here don't close it

`decryptSecretEdit` validates that the target id is present and that the decrypted payload's key matches the envelope's target, then does `protocol.Key = target` and returns `&events.Message{Info: evt.Info, Message: decrypted}`.

Both checks are about **target consistency**, not **editor identity**. `evt.Info` still belongs to the envelope, so after the re-parse the edit carries the envelope sender, and the upsert keys on chat + message id alone. A participant's envelope targeting another member's message therefore replaces that row's text and sender.

Same hole in my branch, for what it's worth, and worse: I assigned the envelope's `SenderJID` onto the original explicitly.

## Shape of the missing check

Before accepting the mutation, compare the envelope sender against the stored original's sender, on canonical identities (the LID/PN aliasing in this repo makes a raw string compare unsafe), and reject when they differ: warn, leave the row untouched, and do not publish a webhook. A rejection test with two distinct participants in one group would lock it in.

A caveat on my side: I have **not** tested the attack end to end against a live account, so treat the exploitability as reasoned from the sources above rather than demonstrated. The decryption asymmetry itself is in the quoted code.

## Field note, in case it's useful

Reproduced on a real account on `0.17.1` with the daemon connected, so the gap is live in the latest release. Incoming DM, edited 15 seconds later:

```
msg_id  A517346F…   12:29:56   edited 0   text 'Uno'   ← original, unchanged
msg_id  A5F0A6BF…   12:30:11   edited 0   text NULL    ← the envelope, empty
```

```
stored message A5F0A6BF… in <redacted>@s.whatsapp.net
without content: unhandled payload secretEncryptedMessage
```

The edited body is in no row of that chat.

Worth noting for #362/#245: the older diagnosis pointed at `extractProtocolMutation`, whose `MESSAGE_EDIT` case is correct and whose test builds that protobuf directly, so it passes. The edit never reaches it. That is why this looked fixed for weeks while it wasn't, and why this PR's approach of decrypting the envelope first is the right shape.

Happy to help with the authorization check or its test if useful.

steipete:
Landed as ae794d9292353f9c7c84a9663063ed9b0558a4a1 after repairing and validating the existing PR.

The key finding from the real-SDK proof is that successful decryption and authorization are separate: a different participant's correctly encrypted edit decrypts, then is rejected by the application. Forging FromMe/participant instead fails the SDK's sender-scoped secret lookup. LID aliases and newest-first history remain supported without depending on an existing message row.

Review also exposed wrapped-history detection and empty-edit data loss; both were reproduced and fixed with regression tests. The complete repository gate, focused race tests, final isolated P0–P2 review, exact-head CI/Docker, and 38 compiled-CLI scenarios passed. Proof is recorded in the PR body; transport is synthetic and the SDK/store/crypto/webhook paths are real.

Thanks @goutamadwant for the implementation, @dvainrub for the authorization analysis, and @Entretoize for the earlier live wire-format evidence. Contributor credit is retained in the squash commit.

PR #422, opened 2026-09-14
feat(messages): add mark-read to send read receipts for received messages

### LINKED ISSUE #421: Feature: send real read receipts for received messages (blue ticks), e.g. `wacli messages mark-read`
## Summary

There is currently no wacli command that sends a WhatsApp **read receipt** for a received message, so the sender never sees blue ticks.

`wacli chats mark-read` (v0.18.2, `internal/app/chat_state.go` → `wa.(*Client).MarkChatAsRead` → `appstate.BuildMarkChatAsRead`) sends a `MarkChatAsRead` **app-state patch**. That syncs the chat's read flag to the linked account's own devices, which is what the docs describe ("State commands send WhatsApp app-state patches"), but it is not a receipt: the other participant's message stays on gray double ticks.

whatsmeow exposes the receipt itself as `Client.MarkRead(ids []types.MessageID, timestamp time.Time, chat, sender types.JID, receiptTypeExtra ...types.ReceiptType)`. Nothing in wacli calls it for the user.

## Use case

A bot/agent account that receives messages from a human and processes them. The natural feedback for the human is the native read receipt: gray = delivered to the bot, blue = the bot actually consumed it. Today the only per-message signal available through wacli is a reaction, which is noisier and non-native.

(Context: agentwatch — https://github.com/agentwatch-sh/agentwatch — routes WhatsApp group messages into AI coding sessions and marks them read once the agent's own hook proves the text entered its context; it currently falls back to a ✅ reaction because of this gap.)

## Proposal

A message-level command:

```
wacli messages mark-read --chat <JID|name> --id <MSGID> [--id <MSGID>...] [--sender <JID>]
```

- Sends `Client.MarkRead` for the given ids in that chat. `--sender` is required for group messages when the message is not in the local store (same rule as `send react --sender` / `poll vote --sender`); otherwise it is looked up from the store.
- Delegated through the `sync --follow` process like sends/reactions (`cmd/wacli/send_ipc.go`), so it works while continuous sync owns the store.
- Honours `--read-only` (it writes to WhatsApp).
- JSON output: `{"chat": ..., "ids": [...], "sent": true}`.

Optionally a `--receipt` flag on `chats mark-read` that additionally sends receipts for the chat's unread messages; but a per-message command is the minimal, explicit primitive.

Notes for implementation: in groups the receipt must be addressed to the participant (`sender`) with `chat` = the group JID; WhatsApp always delivers read receipts in group chats regardless of the read-receipts privacy setting, so this works for bots that keep receipts off for 1:1 chats.

I'm happy to open a PR for this if the shape above is acceptable.

AsyncAlchemist:
Closes #421

## What changed

Adds `wacli messages mark-read --chat CHAT --id MSG_ID [--id ...] [--sender JID] [--pick N]`, which sends WhatsApp **read receipts** (whatsmeow `Client.MarkRead`) for received messages so their sender sees them as read (blue ticks) when read receipts are on for the account.

This is the per-message counterpart of `chats mark-read`, which sends a `MarkChatAsRead` app-state patch that only syncs the chat's read flag to the account's own linked devices and never notifies the sender.

- `internal/wa`: `MarkRead(ctx, chat, sender, ids) (types.ReceiptType, error)` wrapper on the client (added to the `WAClient` interface + fake). It returns the receipt type WhatsApp actually sent: `read`, or `read-self` when the account's read-receipts privacy setting is `none` (or the chat is a newsletter) — the same substitution whatsmeow makes inside `MarkRead` (`receipt.go:214-217`), which it does not report. The settings are read with `TryFetchPrivacySettings`, whose successful fetch is cached and reused by whatsmeow's own `MarkRead`, so both decisions share one snapshot; a failed fetch fails the command. The privacy setting is never changed.
- `internal/app`: `MarkMessagesRead` — every `--id` the local store knows as one of this account's own outgoing messages (`FromMe`) is rejected, on every path (direct chats, and groups with or without `--sender`). The store is checked under the chat's phone-number and LID forms (same PN⇄LID resolution as `messageChatJIDFilter`). Only `sql.ErrNoRows` counts as "not synced" — such IDs are accepted in direct chats and, with `--sender`, in groups, so an unsynced received message can still be marked read; any other store error fails the command. In a group chat with `--sender` omitted, the participant is resolved from the store (every id must be a stored received message from the same participant; otherwise a clear error asks for `--sender`).
- `cmd/wacli`: the command; delegated as kind `read_receipt` through the same-store `sync --follow` process (like sends/reactions), honoring `--read-only`. JSON output: `{"sent": true, "chat": ..., "ids": [...], "receipt": "read" | "read-self", "sender_notified": true | false}`. For `read-self` the human output prints a stderr note (sender will not see blue ticks); under `--events` that hint is a `warning` NDJSON event (`code: read_self_receipt`) so the event stream stays parseable.
- `docs/messages.md` (new "Read receipts" section + synopsis/example), `CHANGELOG.md` (Unreleased).

## Why

A bot/agent account that consumes messages has no native way to tell the human "I've read this". Reactions are the only per-message signal available today and they are noisy and non-native. See #421 for the motivating use case (agentwatch routes WhatsApp group messages into AI coding sessions and wants to mark them read only once the agent actually consumed them).

## Changes after the ClawSweeper review (rev 2)

- **Privacy guarantee corrected.** The docs/help no longer claim that group receipts are always delivered. The wrapper reports the effective receipt type (`read` vs `read-self`), the result carries `receipt` + `sender_notified`, and the docs describe that a disabled read-receipts privacy setting yields `read-self` in direct and group chats alike. whatsmeow's own privacy handling is preserved unchanged.
- **Own messages rejected before dispatch on every path.** The `FromMe` check moved out of the sender-inference branch; regression tests cover the direct-chat and the explicit-`--sender` bypass paths, plus the unsynced-received path staying accepted.
- Rebased onto current `main` (CHANGELOG conflict).

## Changes after the second ClawSweeper review (rev 3, `15af9a9`)

- **Aliased chat identities**: own-message lookup now also checks the chat's mapped LID / phone-number key, so a `FromMe` row stored under the alias is rejected (tests: stored-under-PN with `--chat` as LID, and the reverse).
- **Store errors propagate**: only `sql.ErrNoRows` activates the unsynced-message allowance; any other `GetMessage` error fails the command before dispatch (test: closed store → error, no `MarkRead` call).
- **One privacy snapshot**: the wrapper uses `TryFetchPrivacySettings` (cached on success, reused by whatsmeow's `MarkRead`) and fails on a fetch error instead of reporting `read` from an empty struct.
- **`--events`**: the read-self hint is emitted through `out.EventWriter` as a `warning` event; plain stderr note only without `--events`/`--json` (tests cover all three output modes).

## How it was tested

Full gate: `gofmt` clean, `go vet ./...` clean, `pnpm lint:deadcode` clean, `go test ./...` and `go test -tags sqlite_fts5 ./...` green, `go build -tags sqlite_fts5 ./...`, `pnpm test:docs` (26 pass / 0 fail), `pnpm docs:site`, `git diff --check`.

Unit tests: `internal/app/receipts_test.go` (sender resolved from the store; missing-store hint; explicit sender + blank ids; own message rejected in a direct chat; own message rejected in a group with explicit `--sender`; unsynced direct-chat id accepted; `read-self` passed through from the client) and `cmd/wacli/send_ipc_test.go` (delegate routing/validation; the executor passing chat/ids/sender and the receipt type back in the response).

### Live runtime evidence (rev 3 build `15af9a9`, redacted)

Account: a dedicated agent WhatsApp number with **read receipts ON** (the `receipt: "read"` below is the wrapper reading that setting from whatsmeow's privacy settings). A `wacli sync --follow --presence-mode quiet` process owns the store, so the command is delegated to it over the private socket. `<group>` is a 2-member group (agent account + the human); `<human>` is the human's number.

```console
$ wacli messages mark-read --chat <group>@g.us --id 3AB774A901607987CBC5 --json
{"success":true,"data":{"chat":"<group>@g.us","ids":["3AB774A901607987CBC5"],"receipt":"read","sender_notified":true,"sent":true},"error":null}

$ wacli messages mark-read --chat <group>@g.us --id 3AB774A901607987CBC5
Marked 1 message(s) as read in <group>@g.us

$ wacli --events messages mark-read --chat <group>@g.us --id 3AB774A901607987CBC5
Marked 1 message(s) as read in <group>@g.us
(no stderr output: with read receipts on there is no hint; a read-self result emits {"event":"warning","data":{"code":"read_self_receipt",...}} instead of a plain line)

$ # one of the agent account's own messages -> rejected (no --sender)
$ wacli messages mark-read --chat <group>@g.us --id 3EB0CC929D4BD07BEB8786 --json
{"success":false,"data":null,"error":"message 3EB0CC929D4BD07BEB8786 was sent by you; read receipts apply to received messages"}

$ # same, with --sender (the explicit-sender path) -> still rejected
$ wacli messages mark-read --chat <group>@g.us --id 3EB0CC929D4BD07BEB8786 --sender <human>@s.whatsapp.net --json
{"success":false,"data":null,"error":"message 3EB0CC929D4BD07BEB8786 was sent by you; read receipts apply to received messages"}

$ wacli --read-only messages mark-read --chat <group>@g.us --id 3AB774A901607987CBC5 --json
{"success":false,"data":null,"error":"read-only mode: command would intentionally modify WhatsApp or the local store"}
```

`3AB774A901607987CBC5` is a photo the human sent to the group. Sender-side evidence from the human's WhatsApp (agent account's number redacted; the photo is a public egg-carton label): the message shows blue ticks and the "Seen by" panel lists the agent account. The seen time (2:01 AM local = the send time) is the **first** receipt for that message, sent by the rev-1 build of this same command through the same delegate path right after it arrived; WhatsApp keeps the first read time, so the rev-2/rev-3 runs above re-sent a receipt without changing it.

Blue ticks and Seen-by panel on the human's phone

The `read-self` branch, the aliased-chat rejection and the store-failure path are covered by unit tests only: `read-self` needs read receipts turned off on the agent account's phone, and the other two need an aliased row / a failing store, none of which this live account produces.

## New flags

`messages mark-read`: `--chat`, `--id` (repeatable), `--sender`, `--pick`. No new env vars. The delegate request gains an `ids` field and the response a `receipt` field (additive; older follow processes answer "unsupported send kind" and should be restarted after upgrading, same as the 0.18.2 mark-read delegation note).


AsyncAlchemist:
Rev 2 pushed (`8e915a6`, rebased onto current `main`). Addressing the two findings:

**Unconditional group-receipt guarantee** — confirmed against the pinned whatsmeow (`receipt.go:214-217`): `MarkRead` swaps `read` for `read-self` whenever `ReadReceipts == none`, with no group exception, so the doc line was wrong. Fixed by reporting rather than promising: the `internal/wa` wrapper now returns the effective `types.ReceiptType` (evaluating the same rule whatsmeow applies internally, since `MarkRead` does not report it), the command result carries `receipt: "read" | "read-self"` and `sender_notified`, the human output prints a note for `read-self`, and the docs/help describe the privacy dependence. whatsmeow's privacy handling is untouched; the setting is never changed.

**Own messages reaching `MarkRead`** — confirmed: the `FromMe` check lived only in `storedSender`, so direct chats and groups with `--sender` bypassed it. `MarkMessagesRead` now looks every id up in the store independently of sender inference and rejects any stored `FromMe` message on all paths; ids not in the store stay accepted in direct chats and with an explicit `--sender` (unsynced received messages). Regression tests cover the direct-chat bypass, the explicit-sender bypass, and the unsynced-accepted case.

Redacted live output from the rev-2 build (delegated through `sync --follow`, read receipts ON on the agent account → `receipt: "read"`, own-message rejection on both paths, `--read-only` refusal) is in the PR body under "Live runtime evidence". The `read-self` branch is unit-tested only; producing it live requires turning read receipts off on the agent account's phone.

AsyncAlchemist:
Sender-side evidence (the human's WhatsApp, agent account number redacted): blue ticks on the photo message `3AB774A901607987CBC5` and the "Seen by" panel listing the agent account. The seen time equals the send time because the first receipt for this message was sent by the rev-1 build of this command right after it arrived, over the same delegate path; WhatsApp keeps the first read time, so today's rev-2 run re-sent a receipt without changing it. Also added to the PR body under "Live runtime evidence".

Blue ticks and Seen-by panel

AsyncAlchemist:
Rev 3 pushed (`15af9a9`). All four findings verified in the code and addressed:

- **Mapped chat identities** — confirmed: `GetMessage(chat, id)` was an exact-key lookup, so a `FromMe` row stored under the chat's LID (or PN) alias slipped through. `MarkMessagesRead` now builds the same PN⇄LID key set `messageChatJIDFilter` uses and checks every key. Tests: stored-under-PN with `--chat` as LID, and the reverse, both rejected with no `MarkRead` call.
- **Database errors** — confirmed: every `GetMessage` error was treated as "not synced". Only `sql.ErrNoRows` does that now; any other error fails the command before dispatch. Test: closed store → error, no `MarkRead` call.
- **Separate privacy snapshot** — confirmed: `GetPrivacySettings` returns an empty struct (→ `read`) on a failed fetch without caching it, so the inner call could then fetch `none` and send `read-self`. The wrapper now calls `TryFetchPrivacySettings(ctx, false)`: a success is cached and whatsmeow's own `MarkRead` reads that cache, so both decisions share one snapshot; a failed fetch is returned as an error instead of a guess.
- **`--events`** — confirmed. The read-self hint goes through `out.EventWriter` as a `warning` event (`code: read_self_receipt`, `chat`, `receipt`) when `--events` is set; the plain `Note:` line only prints without `--events`/`--json`. Tests cover the three output modes.

Full gate green (format, vet, deadcode, plain + FTS tests, build, docs tests/site, diff check). The PR body's live evidence is refreshed from the rev-3 build (same delegated `read` receipt, own-message rejection on both paths, `--read-only` refusal, plus an `--events` run). The three new paths (aliased row, failing store, `read-self`) are unit-tested only; this live account produces none of them.

PR #427, opened 2026-09-18
fix: resolve PN/LID identities in history backfill and contact lookup

ugoi:
Mapped WhatsApp contacts can have a phone-number JID in local storage while the primary phone expects its LID for on-demand history. Backfill currently uses the supplied JID for both storage and the request, then matches the response by literal ID. This can time out despite an online phone; passing the LID instead can fail to find a migrated local anchor. Contact search also returns both stored identities and can expose an opaque LID as a phone number.

This change:
- Resolves local backfill anchors after connection/migration, sends the corresponding LID when available, and matches either verified identity in responses. Existing anchor retries, bounds, groups, and unmapped identities are preserved. Request events include `request_chat_jid`.
- Folds verified PN/LID pairs in `contacts search` and `contacts show`, prefers the real phone JID, leaves unknown LID phones empty, preserves searchable metadata and unions tags, and applies limits after folding. Equal names alone never merge contacts; raw contact rows are retained.

Validation: regression tests failed before the fixes and pass afterward, including multi-batch and manual history responses, mappings learned during connection, unrelated responses, groups, unmapped IDs, duplicate limit handling, LID-only and PN-only contacts, aliases/tags, and distinct people with the same name. The required full gate passes: formatting, vet, production/test dead-code checks, plain and FTS Go suites, Windows lock cross-compile, CGO-required check, 26 documentation/release tests, build, docs-site build, and whitespace. Targeted backfill race tests also pass.

No new flags, dependencies, schema migrations, or message-body changes. Contact display reads all local contact metadata before limiting; this prioritizes correct merging and searches of names on either stored row.

Live opt-in validation on one paired account: phone-JID backfill received three on-demand responses and recovered the previously missing older history into the canonical chat. Contact search folded 198 verified duplicate pairs, search/show returned the same real phone identity, and all remaining unmapped LID rows had empty phone fields. No outbound chat messages were sent during validation.


steipete:
Landed with the PN/LID history and contact fixes, completed metadata round-trips, and a response-persistence ordering fix. Contributor credit is preserved in the squash commit and changelog.

Proof: the history identity tests fail against pinned main c8fe691 and pass on this change. Command regressions reproduced disappearing alias/tag edits and removals, metadata held under an identity without a contact row, and writes despite a corrupt session store. A controlled asynchronous history response reproduced early backfill termination before persistence. The fixes pass the full required Linux Crabbox gate and targeted backfill race tests; independent P0–P2 review is scoped-clean. Exact head dc17c7c5493ff3219680b0d51a58ee6b0b9152c1 passed CI (test and Linux release builds) and Docker: https://github.com/openclaw/wacli/actions/runs/35690144146.

Maintainer validation used synthetic data, with no live WhatsApp account. Local contact edits retain contact-row counts and session-file bytes.

